PR #12, opened 2017-03-12
Use GameObject transform for GLTFScenes

stevenvergenz:
Previously, the object's transform would be cancelled out, and the scene would always spawn at world origin.


robertlong:
Looks good to me. Thanks!

Issue #6, opened 2017-03-09
Add Support for GLTF Binary Format .glb


robertlong:
https://github.com/KhronosGroup/glTF/blob/2.0/specification/2.0/GLB_FORMAT.md

PR #13, opened 2017-03-13
Binary glTF support

stevenvergenz:
This pull request adds support for the GLB model file format. All the included models load.

Addresses #6

### DIFF
--- a/Assets/GLTF/Scripts/Exceptions.cs
+++ b/Assets/GLTF/Scripts/Exceptions.cs
@@ -1,0 +1,16 @@
+﻿using System;
+
+namespace GLTF
+{
+	[Serializable()]
+	public class GLTFHeaderInvalidException : Exception
+	{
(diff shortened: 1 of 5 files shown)


robertlong:
Thanks @stevenvergenz! This is great. I moved the GLTFInternalBuffer creation to the GLTFParser class. It seemed a bit cleaner since it doesn't add additional members to the GLTFRootRef and the GLTFBufferId class doesn't require any additional logic in its getter.

Also in the process I noticed you changed the requirement property for buffer.uri and found that GLTF's spec didn't have it: https://github.com/KhronosGroup/glTF/issues/872

Thanks again!

stevenvergenz:
Yeah, I wasn't sure what to do about that URI. Ideally it would be REQUIRED for plain JSON files, and DISALLOWED for binary files. Not sure how to do that short of writing a new class, and that seemed like overkill.

PR #14, opened 2017-03-13
Altspace vr feature/binary support

robertlong:
Additional changes to #13

Some style fixes, additional examples, and moved internal buffer creation into the GLTFParser class.


stevenvergenz:
Ah, good call moving the internal buffer swap. I wasn't able to figure out when the Unity objects were created until just now, so I put it in that strange place. And thanks for updating the examples, I use an old 5.4 build on my machines.

PR #20, opened 2017-05-17
Add field to loader: last loaded model

stevenvergenz:
It's strange to me that the loaded scene is discarded entirely if another game object is not given as its parent. This PR allows consumers to get the reference to the model after loading is complete.


robertlong:
I changed LastLoadedModel to LastLoadedScene. I think it better represents the GameObject.

LGTM 👍

PR #29, opened 2017-05-25
Parse base 64 string prefixes other than application/octet-stream

jamesgk:
For example, sometimes the prefix can be "data:image/jpeg;base64,". We can also make the regex a little more strict if you want to only allow certain known prefixes.

### DIFF
--- a/Assets/GLTF/Scripts/GLTFLoader.cs
+++ b/Assets/GLTF/Scripts/GLTFLoader.cs
@@ -1,6 +1,7 @@
 using UnityEngine;
 using System.Collections.Generic;
 using System.Collections;
+using System.Text.RegularExpressions;
 using System;
 using UnityEngine.Networking;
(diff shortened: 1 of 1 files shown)


Assets/GLTF/Scripts/GLTFLoader.cs
```diff
@@ -406,7 +407,7 @@ private UnityEngine.Material CreateMaterial(Material def, bool useVertexColors)
 			return material;
 		}
 
-		private const string Base64StringInitializer = "data:application/octet-stream;base64,";
+		private const string Base64StringInitializer = "data:[a-z-]+/[a-z-]+;base64,";
```
stevenvergenz: Force this to only match the start of the string, for performance reasons. `^data:`

jamesgk:
Good call, PTAL.

PR #37, opened 2017-07-23
Custom request headers and other fixes

vsantosu:
This pull request includes additional features and some fixes related to [[36](https://github.com/AltspaceVR/UnityGLTF/issues/36) and [35](https://github.com/AltspaceVR/UnityGLTF/issues/35)].

New Features:

1- Optional Request Headers Dictionary parameter for **GLTFLoader** class. If provided, it will add those headers to the GTLF http request. This is required if GLTF assets are on a secured server.

2- Optional headers can be added via the GLTFComponent UI as a list.

3- Added a new Scene **BoomBoxTestScene_secure** to test custom header.

Fixes:

1- Added URLEncoding to all URI found in the GTLF header file(external textures, and buffer). If a texture or the buffer contains illegal characters, the request will fail. This fix should be enough.

2- Added a conditional loop to GLTFComponent to prevent race condition between WebServer and loader.

I tried to follow the author's coding style, if any other change is needed please let me know. Thanks again for your contribution!


vsantosu:
Since there is a major architecture change, this Pull request may not be necessary anymore. I needed to add header options on the Unity HTTP(s) request for authenticated backend, but I guess it is better to implement this in the new project organization. Also, the URL request where not enconded, not sure if the bug is still there.

Assets/GLTF/Scripts/GLTFComponent.cs
```diff
@@ -18,9 +29,22 @@ class GLTFComponent : MonoBehaviour
 
 		IEnumerator Start()
 		{
+			/* prevent race condition with web server */
+			while (webServer == null)
```
MSblgross: Add a comment for why this code is waiting on webserver/expected scenario

Assets/GLTF/Scripts/GLTFComponent.cs
```diff
@@ -1,14 +1,25 @@
...
 using UnityEngine;
 using System.Threading;
+using System.Linq;
 using UnityEngine.Networking;
 
 namespace GLTF {
 
 	class GLTFComponent : MonoBehaviour
 	{
+		[Serializable]
```
MSblgross: Why is this Serializable?

MSblgross:
I think this will be helpful to pull in with the architecture change. I will pull these changes in to that refactor. The API surface will change a bit.  Everything looks good. I left a few comments. The TriangleWithoutIndicesTestScene now fails however. That should be fixed before checkin.

Assets/GLTF/Scripts/GLTFComponent.cs
vsantosu: This was a problem with OSX and the build in web server to test the models. I recall one pull request fixing this issue, so it may not be longer necessary.

Assets/GLTF/Scripts/GLTFComponent.cs
vsantosu: If I remember correctly, the struct needed to be serializable in order apply reflection in the consequent operations, this may not be needed at all. Not remember for sure why I did.

blgrossMS:
This has been hanging a while. I'm going to close the review and open an issue to make sure that we have the ability in our WebRequestLoader to accept headers.

Issue #10, opened 2017-03-09
Improve JSON Deserialization Performance


robertlong:
Currently JSON deserialization is a bit slow and not concurrent. It blocks the renderer and causes dropped frames. I'd like to move it to another thread or at least make it concurrent. There's also likely a lot of room for improvement in overall parsing speed.

robertlong:
Performance greatly improved via manual deserialization in https://github.com/robertlong/UnityGLTFLoader/commit/db40b19589113500becf90650c95d755c9e54b48

Previously we were using JSON.NET's [JsonConvert.DeserializeObject](http://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_JsonConvert_DeserializeObject__1.htm) which uses reflection to deserialize the provided object.

I've moved everything to manual deserialization which removes all reflection and really speeds things up.

Sample Model | Reflection Parse Time | Reflection GCAlloc | Manual Parse Time | Manual GCAlloc |
------------ | ------------- | ---------- | ---------- | ----------
Lantern | 389.40ms | 7.6MB | 31.56ms | 99.6KB |
Corset | 377.99ms | 7.8MB | 29.36ms | 55.7KB |
BoomBox | 384.86ms | 7.8MB | 29.29ms | 85KB |

There's probably still a lot more room for performance improvements, but I'm feeling pretty good about our parsing performance now.

stevenvergenz:
Wow, that's some intense savings. Nicely done!

donmccurdy:
Any idea why parsing here would be slower than JSON parsing in e.g. the browser? All three of the sample models above take ~3ms with `JSON.parse()` in Firefox, or 10x faster.. I suppose that's partly because the C# is processing the model into structured objects and typed values, whereas JS is looser about this?

PR #39, opened 2017-08-19
Architecture and functional changes to gltf repo to divide into components

MSblgross:
[/] Created a .NET schema parsing project and a Unity project to have as separation between parsing and scene construction. This involved moving around files to various directories
[-] Removed all data storage from schema layer. Moved the data storage to be in unity layer
[+] Added in support to load all data from file streams
[+] Added support for building on the UWP platform
[/] Added unit test for .NET schema parser
[/] Refactored API names and moved position of API's to make logic clearer


stevenvergenz:
Wow, this is a huge PR! Kind of hard to review something that touches every file in the project. Could you explain some of these to me? Specifically:

1. Scene construction and schema parsing were already separated, weren't they? All the construction work happened in GLTFLoader, and the schema classes were only for de/serializing and storage.
2. If I understand your second bullet, the schema classes are now strictly de/serializing, and the data is stored in Unity? So there's no intermediate data structure anymore?

MSblgross:
Hey Stephen,

I emailed Robert about the change last night, we had also talked a bit before about the changes that I am proposing, whatever is the best way to review the changes I am up for. The architecture changes are separating them out into a dll that simple does schema parsing, and having the scene construction be what is in the Unity layer (and all the memory for scene construction living there too). The data structure now for scene construction is in the unity layer.

I can separate my changes into a three (or more) PR’s:

  1.  The change of moving the parsing code to its own C# project/dll
  2.  The proposed naming changes we would like to make
  3.  The additional changes made on top of it

Does that sound good?

The reasoning behind these changes are:

  1.  We wanted to be able to handle the parsing of the glTF JSON in an engine abstract layer
  2.  We wanted to be able to control when things are loaded/parsed

The changes I am planning on making still are:

  1.  Adding some more API’s to the Unity layer to allow more modular scene construction (so that we can request that the library just builds a mesh, but then we have our own layer that builds the materials)
  2.  Doing an optimization pass
  3.  Work on extras serialization

-Blake

robertlong:
Thanks for making this PR Blake! There's a lot of good stuff in here. Excited to see changes that bring UWP support.

I'm a big fan of splitting out the JSON serialization/deserialization into a separate DLL. There's really no need to have that in the Unity project and a lot of benefit to those using C#.

@bghgary commented in our email exchange about using [glTF-CSharp-Loader](https://github.com/KhronosGroup/glTF-CSharp-Loader) for serialization/deserialization. It has some benefits like generating its parser from the glTF JSON Schema, but it's also considerably slower because it generates classes and then uses JSON.NET's reflection based parsing. I did some testing on the reflection based parsing in https://github.com/KhronosGroup/UnityGLTF/issues/10 and I don't think we should return to that method. However, we could modify glTF-CSharp-Loader to use the manual JSON.NET serialization/deserialization that we have in UnityGLTF. I'm starting to think that this is the best path forward. It would give a huge speed boost to anyone already using the C# loader, UnityGLTF could focus solely on Unity related things and we could avoid duplicated efforts.

I'm happy to put in the work to make this happen if we decide this is a good path forward.

robertlong:
One more thing to figure out is Khronos and Vendor e

MSblgross:
Modifying the glTF CSharp Loader to support manual serialization/deserialization sounds fine to me. So would the work flow then be:

1) You do the work to change the architecture of the codebase to utilize the loader
2) Once that is complete I'll merge the changes and then send out a new PR of the changes that I have made to the Unity project?

MSblgross:
PR was split up into PR's #44 and #41

PR #44, opened 2017-09-05
Project compiling with architecture changes

MSblgross:
[+] Added platform agnostic math classes
[-] Removed caching from serialization layer
[+] Added caching to unity layer via assetcache object
[+] Added tests around gltf serialization
[+] Added ability to have assets "streamed" from disk instead of loaded via uri

Next PR will have changes that pass around streams instead of byte arrays.


robertlong:
Overall it looks good, all the examples work.

I had some errors when building GLTFSerialization though:
```
Microsoft.NETCore.Portable.Compatibility 1.0.2 provides a compile-time reference assembly for mscorlib on UAP,Version=v10.0, but there is no run-time assembly compatible with win10.
```
and
```
Some packages are not compatible with UAP,Version=v10.0.
```

I found that commenting out the `win10` build target in [GLTFSerializationUWPTests/project.json#L11](https://github.com/MSblgross/UnityGLTF/blob/architecture-update/GLTFSerialization/Tests/GLTFSerializationUWPTests/project.json#L11) resolved the issue.

Also, [GLTFJsonSerialization.dll.mdb](https://github.com/MSblgross/UnityGLTF/blob/architecture-update/UnityGLTF/Assets/GLTF/Plugins/GLTFJsonSerialization.dll.mdb) should no longer be needed.

PR #56, opened 2017-09-14
Fix for bug where matrix was row major instead of column major

MSblgross:
~ Made matrix4x4 column major instead of row major
~ Added in integration test for buggy asset
~ Added test scenes for buggy (which uses matrix)
- Deleted binary files. We may want to check these in, but they should be built by a build server.


robertlong:
I'm seeing a bunch of errors in the model on Mac 2017.1.1f1:
screen shot 2017-09-19 at 10 53 57 am

These errors don't seem to exist on master:
screen shot 2017-09-19 at 10 43 24 am

ThreeJS for comparison:
screen shot 2017-09-19 at 10 44 35 am

Are you seeing something different on your branch and master?

MSblgross:
Good find. I think I've root caused why this is happening. It's a different issue. I'll update this PR with the fix to it as well.

PR #71, opened 2017-09-28
Fix for UWP not building properly

MSblgross:
* Fixes issue where UWP does not build properly due to GLTFComponent issue
* Also am checking in meta files so that users do not have to reconfigure them
* Updated readme to explain build process

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,11 @@ The library will be modularized such that it can be extended to support addition
 Work Items and Issues targeting a 1.0 release of the library can be found in
 [Road to 1.0](https://github.com/KhronosGroup/UnityGLTF/projects/1)
 
+## Building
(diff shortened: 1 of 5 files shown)


andybak:
After pulling the the fix above I'm getting build failures only on Linux and MacOS:

```
Build failed with Exception <<<<
BuildPipeline.BuildPlayer() returned: "Errors: Packaging assets - resources.assets
Shader error in 'GLTF/PbrSpecularGlossiness': syntax error: 
unexpected token 'vi' at 
/github/workspace/Library/PackageCache/org.khronos.unitygltf@99c96d9029/Runtime/Shaders/VertexColor.cginc(28)
```

andybak:
@hybridherbst Let me know if you'd prefer me to open a new issue.

PR #74, opened 2017-10-07
KHR_materials_pbrSpecularGlossiness support

jm991:
There are quite a few areas where this could be improved, but it does load spec/gloss materials 

Areas that still need work:

- [x] Add a test to GLTFSerialization project for the KHR_materials_pbrSpecularGlossiness test file I added (1501770225815.gltf)
- [x] Add toggle to GLTFComponent.cs that allows for optional addition of collision geometry and layer
- [x] Revert editor downgrade
- [x] Clean up test scene & reference approved spec gloss gltf file from [GLTF samples](https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/Lantern/glTF-pbrSpecularGlossiness/Lantern.gltf)
- [x] Double checked the full spec of [KHR_materials_pbrSpecularGlossiness schema](https://github.com/KhronosGroup/glTF/blob/master/extensions/Khronos/KHR_materials_pbrSpecularGlossiness/schema/glTF.KHR_materials_pbrSpecularGlossiness.schema.json) to make sure all properties are properly implemented in the Serialize/Deserialize methods 
- [x] Removed PbrSpecularGlossiness GLTFProperty in favor of KHR_materials_pbrSpecularGlossiness Extension in GLTFSerialization project
- [x] in GLTFSerialization project Material.cs's case "extensions" might not be the best way to handle material extensions (is there a better way to iterate through the material extensions than the while loop I added?)


GLTFSerialization/GLTFSerialization/Schema/Material.cs
```diff
@@ -132,9 +156,15 @@ public static Material Deserialize(GLTFRoot root, JsonReader reader)
...
+            }
+
+            if (PbrSpecularGlossiness != null)
+            {
+                writer.WritePropertyName("pbrSpecularGlossiness");
+                PbrSpecularGlossiness.Serialize(writer);
+            }
 
-			if (CommonConstant != null)
+            if (CommonConstant != null)
```
MSblgross: nit- make spaces tabs

GLTFSerialization/GLTFSerialization/Schema/Material.cs
```diff
@@ -92,7 +98,25 @@ public static Material Deserialize(GLTFRoot root, JsonReader reader)
 					case "pbrMetallicRoughness":
 						material.PbrMetallicRoughness = PbrMetallicRoughness.Deserialize(root, reader);
 						break;
-					case "commonConstant":
+                    case "pbrSpecularGlossiness":
+                        material.PbrSpecularGlossiness = PbrSpecularGlossiness.Deserialize(root, reader);
+                        break;
+                    case "extensions":
```
MSblgross: extensions are handled by the default case. You should let that process the extensions and hook up an extension handler  via extension factory API.

MSblgross: Also why are there two ways of processing KHR_materials_pbrSpecularGlossiness

PR #90, opened 2017-11-06
Fix for 2017.2

### LINKED ISSUE #58: Current code not working with Unity 2017.2.0b11
I am trying to open the master project in Unity 2017.2.0b11 and I am getting error with the UnityStandardCore.cginc file missing a MetalRough reference.

sbtron:
Potentially fixes #58 by adding  metallicRough to glTFStandardinput.
Disclaimer: I have no idea why this works but it did with 2017.2 :).

I haven't tested this in 5.6 and not sure if an ifdef needs to be added?

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Shaders/GLTFStandardInput.cginc
+++ b/UnityGLTF/Assets/UnityGLTF/Shaders/GLTFStandardInput.cginc
@@ -154,7 +154,6 @@ half4 SpecularGloss(float2 uv)
 	return sg;
 }
 
-#ifdef UNITY_2017
 half2 MetallicRough(float2 uv)
 {
 	half2 mr;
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Shaders/GLTFStandardInput.cginc
```diff
@@ -154,6 +154,17 @@ half4 SpecularGloss(float2 uv)
 	return sg;
 }
 
+half2 MetallicRough(float2 uv)
```
stevenvergenz: Maybe wrap this definition in an `#if UNITY_VERSION >= 2017.2`? (BTW I don't know what the actual version value is since they changed the naming scheme.)

sbtron:
@stevenvergenz I think this should do it. would be great if you confirm too? Documentation seems hard to find on the exact pattern

stevenvergenz:
I still get the shader error with those guards. But you're right, there's no documentation for what the shader macro is for the new versioning scheme. So I guess revert that commit, it does more harm than good. Sorry for the goose chase.

Issue #85, opened 2017-10-20
Exported Model not compliant with specs


vsantosu:
Hello all,

I'm trying to use the exported scene from the test examples and I'm encountering several error. Most of them with offsets, not sure if the loaders(babylon, Threejs, etc) are wrong or it is the exporter. The following is the GLTF model generated using the **ExportTestScene**:

[TestScene.zip](https://www.dropbox.com/s/m8hl30rxc20dsc2/TestScene.zip?dl=0)

Then, if you drag and drop this into babylon.js Sandbox, it shows an offset error:

[Babylon Sandbox](http://sandbox.babylonjs.com/)

It works under Threejs tho:

[ThreeJs Loader](https://gltf-viewer.donmccurdy.com/)

Finally, if you try to validate into Kronos GLTF online tool, it shows several errors:

[Kronos GLTF Validator](http://github.khronos.org/glTF-Validator/)

Thanks so much for your help, I really appreciate it!

sbtron:
The exporter seems broken at the moment. I ran into the same issue and was starting to look into it.

sbtron:
Looks like the original asset itself has issues and the exporter just passes it through.
@robertlong @stevenvergenz  any idea how [TestScene.gltf](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/www/TestScene/TestScene.gltf) was generated?

robertlong:
TestScene.gltf is created by using the exported on [ExportTestScene.unity](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/Assets/UnityGLTF/Examples/ExportTestScene.unity). TestScene.gltf probably shouldn't be checked into the repository.

The export errors are something I'm planning on looking into in a few more weeks. I haven't had the time to work on this project recently.

PR #91, opened 2017-11-07
Fix export

sbtron:
Export is broken due to missing names for textures and images. Using texture count to give all images an unique default name.


sbtron:
Looks like Issue #85 is something else and not related to the core export being broken. Merging this to get export working and will do a separate PR for the other issue.

Issue #101, opened 2018-01-03
Vertical UV coordinate negated not flipped


amenzies:
Hi,

I think I have discovered an issue when reading in a model that a clamp wrap mode.  The issue is that when reading in UV coordinates the vertical component is negated not flipped as seen here:

```
public Vector2[] AsTexcoordArray(ref NumericArray contents, byte[] bufferData)
{
	if (contents.AsTexcoords != null) return contents.AsTexcoords;

	var arr = AsVector2Array(ref contents, bufferData);
	for (var i = 0; i < arr.Length; i++)
	{
		arr[i].Y *= -1;
	}

	contents.AsTexcoords = arr;
	contents.AsVec2s = null;

	return arr;
}
```

I think that the correct approach is to do this `arr[i].Y = 1 - arr[i].Y;`.  I've attached a modified version of the Avocado example mesh from the gltf-sample-models repo.  I've edited it to use a clamp warp mode.
[Avocado.zip](https://github.com/KhronosGroup/UnityGLTF/files/1599232/Avocado.zip)

Note that without code above the avocado renders as a just a smear of color.  With the adjustment it is textured correctly.

I think GLTFSceneExporter may also need to be updated since `InvertY` also multiplies by -1.  Perhaps the function should be updated to apply `1-y` instead and be renamed to something like `FlipY` to avoid confusion with the other invert methods.

For reference https://github.com/KhronosGroup/UnityGLTF/issues/75 covers a similar but different issue.

Cheers!

amenzies:
I've created a pull request https://github.com/KhronosGroup/UnityGLTF/pull/102 to address this issue.

amenzies:
As a side note, I noticed that when using the GLTF->Export Selected option on the Avocado.zip mesh above, the Metallic Roughness texture is exported as the primary texture and neither the albedo or normal map's are exported.  Not related to this bug or it's fix but thought I should call it out.

blgrossMS:
This should be merged in and fixed, thanks for the contribution!

PR #102, opened 2018-01-03
Flip vertical compontent of UVs instead of negating them

amenzies:
Adjustment to how UV.Y is flipped so that clamp warp modes are supported.  See issue https://github.com/KhronosGroup/UnityGLTF/issues/101


MSblgross:
May want to add your sample to the examples. Thanks for fixing this!

PR #103, opened 2018-01-03
Compute stream length relative to position

amenzies:
I'm working on a 3D Tiles renderer for unity.  The 3D Tiles spec wraps glb files in [b3dm](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel) format which is basically just a glb file with some additional header information.  

When GLTFSceneImporter.Load is called it currently uses the entire length of the input stream, which means I need to create an extra memory stream and copy the glb data into it for load to work.  It's desirable to just scrub past the header and then pass in the original stream.  This can be supported by measuring the remaining stream length relative to the streams current position.  This should be fine since the glb element is always the last thing in the b3dm file.

Cheers


MSblgross:
Thanks for the bug fix!

PR #106, opened 2018-01-17
Adding Unity 2017.2 and 2017.3 support, as well as additional UWP support

keveleigh:
This PR adds a few newly generated folders for both UWP and Unity 2017 to the gitignore.

It also add versioning `#if` statements where APIs have changed between Unity versions.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -189,13 +189,12 @@ namespace UnityGLTF
 						yield return LoadImage(_gltfDirectoryPath, image, i);
 					}
 				}
-#if !WINDOWS_UWP
+
 				// generate these in advance instead of as-needed
 				if (isMultithreaded)
 				{
 					yield return _asyncAction.RunOnWorkerThread(() => BuildAttributesForMeshes());
 				}
-#endif
 			}
 
 			var sceneObj = CreateScene(scene);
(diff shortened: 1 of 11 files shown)


MSblgross:
Thanks for the contribution. I should be committing a change soon that does some of the work to upgrade. As a note, you don't want the IntegrationTest utilities in a version past 5.6 since Unity changed how integration tests are run,

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -774,7 +781,11 @@ protected virtual IEnumerator LoadBuffer(string sourceUri, GLTF.Schema.Buffer bu
 				{
 					var www = UnityWebRequest.Get(Path.Combine(sourceUri, uri));
 
+#if UNITY_2017_2_OR_NEWER
```
StephenHodgson: @keveleigh pretty sure this was introduced in 2017.1, but I may be wrong.
Please double check.

StephenHodgson:
@keveleigh there wasn't any shader changes required for 2017.3?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
keveleigh: Double checked, and 2017.1 does not have `SendWebRequest`

image

blgrossMS:
I merge these changes in manually to master. Sorry it took so long, thanks for the contribution!

Issue #109, opened 2018-01-27
implement AnimationClip


ousttrue:
Hello.

I'm working on AnimationClip importing.
There is something I don't understand.
Tell me how to ```GetArray<float>(sampler.Input);``` in below code.

```cs
		AnimationClip ImportAnimationClip(Transform root, Transform[] nodes, int animationID)
		{
			var clip = new AnimationClip();
			clip.name = string.Format("anmation:{0}", animationID);

			var animation = _root.Animations[animationID];
			foreach(var y in animation.Channels)
			{
				var node = nodes[y.Target.Node.Id];
				var relativePath = RelativePathFrom(node, root);
				switch (y.Target.Path)
				{
					case GLTFAnimationChannelPath.translation:
						{
							var curveX = new AnimationCurve();
							var curveY = new AnimationCurve();
							var curveZ = new AnimationCurve();

							var sampler = animation.Samplers[y.Sampler.Id];
							var input = GetArray<float>(sampler.Input); // <- how to ?
							var output = GetArray<Vector3>(sampler.Output); // <- how to ?
							for (int i = 0; i < input.Length; ++i)
							{
								var time = input[i];
								curveX.AddKey(time, output[i].x);
								curveY.AddKey(time, output[i].y);
								curveZ.AddKey(time, -output[i].z); // reverse-z
							}

							clip.SetCurve(relativePath, typeof(Transform), "localPosition.x", curveX);
							clip.SetCurve(relativePath, typeof(Transform), "localPosition.y", curveY);
							clip.SetCurve(relativePath, typeof(Transform), "localPosition.z", curveZ);
						}
						break;

				}

			}

			return clip;
		}
```

stevenvergenz:
There are a couple problems here, all on our end:

* Animation channels' `sampler` fields are mis-typed. It should point to an animation sampler and not a texture sampler. You probably won't be able to correctly dereference it until that's fixed.
* Normally you would call `sampler.Input.Value.AsFloatArray()`, but `AsFloatArray` is not implemented yet. See [the other conversion functions](https://github.com/KhronosGroup/UnityGLTF/blob/master/GLTFSerialization/GLTFSerialization/Schema/Accessor.cs).

If you're looking to implement animations in UnityGLTF, those are great places to start. Please open a pull request once you have something!

ousttrue:
Thanks for your advice.
I made it !

blgrossMS:
This just got merged in, thanks again for the contribution!

PR #113, opened 2018-01-30
Animation importing #109

ousttrue:
I implemented animation import.
And add AnimatedCube example scene from https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedCube


ousttrue:
Sorry, mistake. 
I did not intentionally change the line feed code.
I will do it again.

PR #115, opened 2018-01-31
Fixing loading from StreamingAssets

keveleigh:
This PR re-adds `LoadType.Stream` to GLTFSceneImporter.Load (removed in #104). Otherwise, attempting to lead from StreamingAssets by checking the "Use Stream" box on a GLTF Component throws the "Invalid load type specified" Exception, even though `LoadType.Stream` is supported.

It also trims the beginning of the passed in filename, due to a quirk in `Path.Combine`, found in the way some of the example scenes, like UWPTestScene, specify the filename.

This PR also includes a .meta file that was missing and updates a normal map to be properly denoted as a normal map in Unity.


blgrossMS:
Closing this, changes got included in another PR, and I will send out a PR that adds a new type of loader

Issue #127, opened 2018-02-14
Bug with exporting meshes


DDovzhenko:
class - ASsets/UnityGLTF/Scripts/Extensions/SchemaExtensions.cs
method - ConvertVector3CoordinateSpaceAndCopy
error - when copying to new array wrong values from old are used:

`returnArray[i].x = array[i].x * coordinateSpaceCoordinateScale.X;`
`returnArray[i].`**y**` = array[i].`**z**` * coordinateSpaceCoordinateScale.Y;`
`returnArray[i].`**z**` = array[i].`**y**` * coordinateSpaceCoordinateScale.Z;`

PR #129, opened 2018-02-20
Fixed exporter to export the correct vectors.

justinctlam:


justinctlam:
This should address issue #127

sbtron:
Thanks for the fix!

Issue #131, opened 2018-02-23
Scripted Importer?


lmontoute:
Hey folks!

First off, huge appreciation for the work here to get glTF in Unity! We're using part of this repo for an editor plugin ([thinksquirrel/chipmunk-unity](https://github.com/thinksquirrel/chipmunk-unity)) and added native import in the editor using the [ScriptedImporter](https://docs.unity3d.com/ScriptReference/Experimental.AssetImporters.ScriptedImporter.html) API:

![image](https://user-images.githubusercontent.com/1100891/36618749-ee60528c-18b1-11e8-8e52-60ab3cbc9e44.png)

Would this be of interest at all upstream? If so, I'd be happy to clean it up, base it off of the latest upstream and add a PR. The downside is that this would only supported in Unity 2017 or newer.

sbtron:
This would be greatly appreciated!

PR #132, opened 2018-02-24
ScriptedImporter support

lmontoute:
(also see #131)

Adds a ScriptedImporter for Unity 2017.1 and up, with support for a lot of standard model importer options:

image

Everything is imported as it is at runtime (except the coroutine is run synchronously), then textures are encoded (as png or jpg), and references are re-linked to assets on disk.
 
Only .glb files are supported (Supporting .gltf would be tricky as we'd need to support importing the mesh data separately, as well as linking existing textures - it's also a bit awkward when compared to other Unity model assets).

Some things that are missing (nice-to-haves, really):

- Automatically switch previously imported textures when enabling/disabling "Use JPG textures" (maybe, I would hesitate to delete any files without user confirmation). Existing textures/materials _will_ automatically update with edits to the glTF file, however.
- Automatic tangent generation (though that might be out of scope/too complex right now, unfortunately Unity doesn't expose its tangent generation functionality last I checked)

Feedback is appreciated! Note that I included #115 here. Noticed quite a few pending changes on the loader side of things, but that was needed in order to load from a stream (rather than use unnecessary UnityWebRequests in the editor).

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
@@ -1,0 +1,388 @@
+#if UNITY_2017_1_OR_NEWER
+using UnityEditor.Experimental.AssetImporters;
+using UnityEditor;
+using System.IO;
+using UnityEngine;
+using System.Collections.Generic;
+using System.Linq;
+using System;
+using Object = UnityEngine.Object;
+using System.Collections;
+
+namespace UnityGLTF
+{
+    [ScriptedImporter(1, new[] { "glb" })]
+    public class GLTFImporter : ScriptedImporter
+    {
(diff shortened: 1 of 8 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
```diff
@@ -0,0 +1,388 @@
+#if UNITY_2017_1_OR_NEWER
```
StephenHodgson: Just curious, but what's specific to 2017.1?

lmontoute: `UnityEditor.Experimental.AssetImporters` (scripted AssetImporters) was introduced in 2017.1, I believe (unless it's in a 5.6 patch)

sbtron:
Thanks for the PR!

PR #133, opened 2018-02-27
Fix export of metallic and normal texture map. Add support for any shader that can support metallic roughness workflow.

justinctlam:
Fix export of metallic and normal texture map.  Add support for any shader that can support metallic roughness workflow.


justinctlam:
Some notes about my changes.

I decided to check the properties to determine which material to export instead of using the name is because it will let developers create their own types of shaders that can be used with this system. For example, we are using a shader in the Unity store called Standard Plus which allows for light maps.

justinctlam:
@ziedbha Yes, I took the idea from the SketchFab Unity to GLTF exporter. I thought it would be slow and would think using the GPU to do it would be faster. I haven't had time to figure out how to do it as I'm still learning the Unity APIs. I would love to learn how to do it. Would you be able to update my PR with that enhancement?

ziedbha:
I can definitely do that. I'll sync with your PR and add some changes and it will hopefully boost the speed. I saw you checked PR #134 . To explain, it only fixes color space issues since a lot of textures were not converted in the correct color space using the Blit method. Your suggested changes handle that, but might be really slow. So combining the ability to specify shader properties + speed of getting pixels is the best way to approach this. Therefore I think we should combine both of our changes!

justinctlam:
@ziedbha I agree. Thank you! I can't wait to learn how to do it better.

justinctlam:
Hi, I haven't heard of any feedback other than from @ziedbha. Should I close this PR? Maybe it's not needed?

blgrossMS:
I'll take a look later today. Sorry for the delay

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +125,175 @@ public void SaveGLTFandBin(string path, string fileName)
...
-				Debug.Log(image.name);
-				var renderTexture = RenderTexture.GetTemporary(image.width, image.height);
-				Graphics.Blit(image, renderTexture);
-				RenderTexture.active = renderTexture;
-				var exportTexture = new Texture2D(image.width, image.height);
-				exportTexture.ReadPixels(new Rect(0, 0, renderTexture.width, renderTexture.height), 0, 0);
-				exportTexture.Apply();
-				File.WriteAllBytes(Path.Combine(path, image.name + ".png"), exportTexture.EncodeToPNG());
-			}
+            exportImages(path);
```
blgrossMS: ExportImages is more the C# style and what we have been using in this code base

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +125,175 @@ public void SaveGLTFandBin(string path, string fileName)
...
+                            break;
+                        case IMAGETYPE.A:
+                            inputColor = new Color(inputColor.a, inputColor.a, inputColor.a, inputColor.a);
+                            break;
+                        case IMAGETYPE.G_INVERT:
+                            inputColor = new Color(1.0f - inputColor.g, 1.0f - inputColor.g, 1.0f - inputColor.g, 1.0f - inputColor.g);
+                            break;
+                        case IMAGETYPE.RGB:
+                        case IMAGETYPE.RGBA:
+                            // inputColor = inputColor;
```
blgrossMS: ?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +125,175 @@ public void SaveGLTFandBin(string path, string fileName)
...
+                for (int j = 0; j < textureWidth; ++j)
+                {
+                    int index = i * textureWidth + j;
+                    Color outputColor = outputColors[index];
+                    Color inputColor = inputColors[index];
+                    
+                    switch (inputChannel)
+                    {
+                        case IMAGETYPE.R:
+                            inputColor = new Color(inputColor.r, inputColor.r, inputColor.r, inputColor.r);
```
blgrossMS: Why are you creating a new Vec4 out of a single float field?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +125,175 @@ public void SaveGLTFandBin(string path, string fileName)
...
+        private void exportMetallicGlossTexture(Texture2D texture, string outputPath)
+        {
+            var textureWidth = texture.width;
+            var textureHeight = texture.height;
+            Color[] outputColors = new Color[textureWidth * textureHeight];
+            for (int i = 0; i < outputColors.Length; ++i)
+            {
+                outputColors[i] = new Color(1.0f, 1.0f, 1.0f);
+            }
+            addTexturePixels(ref texture, ref outputColors, IMAGETYPE.R, IMAGETYPE.B);
```
blgrossMS: is the point of this to swap the R and B channel?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +125,175 @@ public void SaveGLTFandBin(string path, string fileName)
...
+                Debug.Log("Failed to convert texture " + texture.name + " (unsupported type or format)");
+                return;
+            }
+            Texture2D newtex = new Texture2D(textureWidth, textureHeight);
+            newtex.SetPixels(textureColors);
+            newtex.Apply();
+            File.WriteAllBytes(Path.Combine(outputPath, texture.name + ".png"), newtex.EncodeToPNG());
+        }
+
+        private void addTexturePixels(ref Texture2D texture, ref Color[] outputColors, IMAGETYPE inputChannel, IMAGETYPE outputChannel)
```
blgrossMS: these functions could all use comments, pretty confused

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -486,11 +679,14 @@ private void ExportTextureTransform(TextureInfo def, UnityEngine.Material mat, s
 			);
 		}
 
-		private NormalTextureInfo ExportNormalTextureInfo(UnityEngine.Texture texture, UnityEngine.Material material)
+		private NormalTextureInfo ExportNormalTextureInfo(
+            UnityEngine.Texture texture, 
+            TextureMapType textureMapType,
```
blgrossMS: use tabs instead of space

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -4,17 +4,46 @@
...
+            RGB,
+            RGBA,
+            R,
+            G,
+            B,
+            A,
+            G_INVERT
+        }
+
+        enum TextureMapType
```
blgrossMS: explicitly declare access modifier

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +125,175 @@ public void SaveGLTFandBin(string path, string fileName)
...
+                        break;
+                }
+            }
+        }
+
+        private void exportMetallicGlossTexture(Texture2D texture, string outputPath)
+        {
+            var textureWidth = texture.width;
+            var textureHeight = texture.height;
+            Color[] outputColors = new Color[textureWidth * textureHeight];
```
blgrossMS: Just make outputColors an out parameter to the addtexturepixels function and have it create the array. The preinitialization with default values shouldn't be needed if you are filling it all in the calling function

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +125,175 @@ public void SaveGLTFandBin(string path, string fileName)
...
+            for (int i = 0; i < outputColors.Length; ++i)
+            {
+                outputColors[i] = new Color(1.0f, 1.0f, 1.0f);
+            }
+            addTexturePixels(ref texture, ref outputColors, IMAGETYPE.R, IMAGETYPE.B);
+            addTexturePixels(ref texture, ref outputColors, IMAGETYPE.A, IMAGETYPE.G_INVERT);
```
blgrossMS: Why don't you swap all the values you want to in a single pass instead of two passes?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Fixed

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Fixed

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: These are removed, now I am using Graphics.blit.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Removed, the channel swap is done in the shader.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: This is removed, now the swap is done in the shader.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: These functions are no longer necessary.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: This function has been removed.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: This is removed.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Should be fixed now.

justinctlam:
I've fixed all the feedback. @ziedbha I've updated to use Graphics.blit. Thanks!

justinctlam:
Hi! Just wanted to see if there are any more issues I can address for this. At the moment exporting GLTF with pbrmaterials aren't working correctly.

UnityGLTF/Assets/UnityGLTF/Scripts/Editor/GLTFExportMenu.cs
```diff
@@ -30,6 +33,8 @@ static void ExportScene()
 
 		var exporter = new GLTFSceneExporter(transforms);
 		var path = EditorUtility.OpenFolderPanel("glTF Export Path", "", "");
-		exporter.SaveGLTFandBin(path, scene.name);
+		if (path != "") {
```
blgrossMS: string.IsNullOrEmpty

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -93,39 +123,105 @@ public void SaveGLTFandBin(string path, string fileName)
...
+					break;
+				default:
+					ExportTexture (image, outputPath);
+					break;
+				}
 			}
 		}
 
-		private SceneId ExportScene(string name, Transform[] rootObjTransforms)
+		private void ExportMetallicGlossTexture (Texture2D texture, string outputPath)
```
blgrossMS: add comments

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -93,39 +123,105 @@ public void SaveGLTFandBin(string path, string fileName)
...
 		{
-			var scene = new Scene();
+			var destRenderTexture = new RenderTexture (texture.width, texture.height, 0, RenderTextureFormat.ARGB32, RenderTextureReadWrite.Linear);
+			var shader = Resources.Load ("MetalGlossChannelSwap", typeof(UnityEngine.Shader)) as UnityEngine.Shader;
```
blgrossMS: cache the shaders at the start of the exporter instead of doing it here

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -93,39 +123,105 @@ public void SaveGLTFandBin(string path, string fileName)
...
+			var metalGlossMaterial = new UnityEngine.Material (shader);
 
-			if (ExportNames)
+			Graphics.Blit (texture, destRenderTexture, metalGlossMaterial);
+
+			var exportTexture = new Texture2D (texture.width, texture.height);	
+			exportTexture.ReadPixels (new Rect (0, 0, destRenderTexture.width, destRenderTexture.height), 0, 0);	
+			exportTexture.Apply ();
```
blgrossMS: I think this will do a CPU and GPU copy even though we just care about the CPU version here to write out. Is there a way to do this without creating an additional texture (like copying the bytes of the destRenderTexture into a file?)

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -93,39 +123,105 @@ public void SaveGLTFandBin(string path, string fileName)
...
+
+			var finalFilenamePath = ConstructImageFilenamePath (texture, outputPath);
+			File.WriteAllBytes (finalFilenamePath, exportTexture.EncodeToPNG ());
```
blgrossMS: This is more a general note for how export to be done than anything that you should do on your review, but we should match the ILoader interface in the importer so that we are using streams in the exporter through a loader like interface and not assuming File I/O or being in control of it here even

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -93,39 +123,105 @@ public void SaveGLTFandBin(string path, string fileName)
...
+					break;
+				}
 			}
 		}
 
-		private SceneId ExportScene(string name, Transform[] rootObjTransforms)
+		private void ExportMetallicGlossTexture (Texture2D texture, string outputPath)
 		{
-			var scene = new Scene();
+			var destRenderTexture = new RenderTexture (texture.width, texture.height, 0, RenderTextureFormat.ARGB32, RenderTextureReadWrite.Linear);
```
blgrossMS: make your rendertextures with gettemporary so that it will potentially pool them: https://docs.unity3d.com/ScriptReference/RenderTexture.GetTemporary.html

blgrossMS: call release at the end of the function

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
blgrossMS: Looking around, it seems like this is the recommended pattern, add a comment explaining why we create another gpu copy of the data

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -93,39 +123,105 @@ public void SaveGLTFandBin(string path, string fileName)
...
+
+			var exportTexture = new Texture2D (texture.width, texture.height);	
+			exportTexture.ReadPixels (new Rect (0, 0, destRenderTexture.width, destRenderTexture.height), 0, 0);	
+			exportTexture.Apply ();
+
+			var finalFilenamePath = ConstructImageFilenamePath (texture, outputPath);
+			File.WriteAllBytes (finalFilenamePath, exportTexture.EncodeToPNG ());
+		}
+
+		private void ExportNormalTexture (Texture2D texture, string outputPath)
```
blgrossMS: comment

UnityGLTF/Assets/UnityGLTF/Scripts/Editor/GLTFExportMenu.cs
justinctlam: Updated

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Added comments

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Caching the material with shaders

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Makes sense. I was just following the existing code already in place.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Cool tip! thanks!

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: Added comments

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: I'm not sure why this is the recommended pattern. Is there a better way to do this? I was following the same pattern as before.

justinctlam:
@blgrossMS I've updated the PR with the feedback you provided. Thanks again for looking at it!
Hi @ziedbha, would appreciated if you can take a quick look at this again.

justinctlam:
Hi, just wanted to see if there are any more issues I should address? Thanks!

blgrossMS:
Sorry I’ve been out at GDC. I’ll take a look this week

blgrossMS:
@justinctlam was testing before merging in, no texture were generated when I did a scene or selection export on the object. I also noticed that an error is thrown if RenderTexture.active is not set to null before release.

justinctlam:
@blgrossMS Thanks for looking, I will investigate those issues.

justinctlam:
@blgrossMS Do you happen to have a test scene I can look at? The scene I'm using exports textures fine for scene and selection object and I don't see any error thrown.

blgrossMS:
I tried the boomboxtestscene and was having issues.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

justinctlam:
@blgrossMS I tried to export a scene and selected object on Master build and it also didn't export the textures.  Is this something my PR should also fix?

blgrossMS:
It does not have to. Can you include the scene that you are testing with in the PR (or is it one of the included ones already) just so that I can verify the functionality?

justinctlam:
@blgrossMS I am using these free pbr assets from the Unity asset store:
https://assetstore.unity.com/packages/3d/props/industrial/pbr-barrels-83508

If you download them, there is a test scene in it.  You can also try out many other free pbr assets on the Unity asset store.  They should all work.

justinctlam:
@RPSpicer I can look into removing the dependency on UnityEditor. If we are okay with naming collision or have an alternative way to resolve it, I'm happy to hear any feedback.

RPSpicer:
@justinctlam Ideally we'd want to know the relative path from the input scene/mesh file to each input texture file, but that can't be guaranteed in all cases, for instance my runtime case (where these were loaded into Unity datastructures using another piece of code and now exist in memory). Since there's so many use cases we may end up needing configuration options to meet all needs.

As a default, I am biased in favor of "name the exported texture using the Unity `Texture2d`'s name (which should be the filename if imported in the editor), and put it next to the exported mesh in the filesystem" because it handles my case.  EG:  `foo.obj` referencing texture `bar.jpg` ends up with `foo.gltf`, `foo.bin`, and `bar.png` in one folder.

 I'd advocate for refactoring Editor specific calls over into the `Scripts/Editor/GLTFExportMenu.cs` script so that the functionality you implemented to find relative paths can be preserved, as well as runtime use. That makes for a more complex call into `GLTFSceneExporter,` since you'd have to pass information about the relative paths to textures in, but isolates the Editor-specific code.  What if we allowed `GLTFSceneExporter.SaveGltfAndBin()` to optionally accept a Dictionary of Texture2Ds to relative paths (strings)? If empty, do the simple/naive thing, if populated, use that to define the relative path for each exported texture.

justinctlam:
@RPSpicer I added a delegate to decouple the UnityEditor from the runtime exporter. Please see the latest updates. Let me know if this would work for you.

RPSpicer:
@justinctlam Thank you! This approach works for me, and enables end users to specify their own arbitrary path logic, which seems like a best case for this library.

PabloMonfort:
sorry to bump over this but, nice tool im try something i need with but there are issues with texture not importing right. i have to export an avatar model to glb thats have a shared material where the textures inputs changes at runtime from some short of list and this dosnt work , even the shoes that has lit urp default not detect the correct textures that has the prefab in runtime, its like if its always pointing to the fbx model or original root of the mesh model and not current materials of the prefab i want to export? sounds confusing can anyone help please ? thanks

PR #134, opened 2018-02-27
Found fix for texture exporting

ziedbha:
I found a fix for the texture exporting, specifically for the bug that had the baseColor be too bright or too dark. The idea is to disable/enable sRGB conversion during the blit operation for each image exported. This is a test run and seems to work, but the most important change I'm working on is to distinguish between Metallic and Base Color (need separate treatment for sRGB for each).

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
@@ -4,29 +4,27 @@ using System.IO;
 using GLTF.Schema;
 using UnityEngine;
 using UnityGLTF.Extensions;
-using UnityEditor;
 
 namespace UnityGLTF
 {
 	public class GLTFSceneExporter
 	{
-        enum TextureMapType
-        {
+		enum TextureMapType
+		{
 			Main,
 			Emission,
 			Occlusion,
 			Bump,
 			SpecGloss,
 			MetallicGloss,
-			Light    
-        };
+			Light
+		};
 
-        struct ImageInfo
-        {
-            public Texture2D texture;
-            public TextureMapType textureMapType;
-			public bool gamma;
-        };
+		struct ImageInfo
+		{
+			public Texture2D texture;
+			public TextureMapType textureMapType;
+		};
 
 		private Transform[] _rootTransforms;
 		private GLTFRoot _root;
@@ -54,9 +52,11 @@ namespace UnityGLTF
 		public GLTFSceneExporter(Transform[] rootTransforms)
 		{
 			_rootTransforms = rootTransforms;
-			_root = new GLTFRoot{
+			_root = new GLTFRoot
+			{
 				Accessors = new List<Accessor>(),
-				Asset = new Asset {
+				Asset = new Asset
+				{
 					Version = "2.0"
 				},
 				Buffers = new List<GLTF.Schema.Buffer>(),
@@ -70,12 +70,13 @@ namespace UnityGLTF
 				Textures = new List<GLTF.Schema.Texture>(),
 			};
 
-            _imageInfos = new List<ImageInfo>();
+			_imageInfos = new List<ImageInfo>();
 			_materials = new List<UnityEngine.Material>();
 			_textures = new List<UnityEngine.Texture>();
 
 			_buffer = new GLTF.Schema.Buffer();
-			_bufferId = new BufferId {
+			_bufferId = new BufferId
+			{
 				Id = _root.Buffers.Count,
 				Root = _root
 			};
@@ -86,7 +87,8 @@ namespace UnityGLTF
 		/// Gets the root object of the exported GLTF
 		/// </summary>
 		/// <returns>Root parsed GLTF Json</returns>
-		public GLTFRoot GetRoot() {
+		public GLTFRoot GetRoot()
+		{
 			return _root;
 		}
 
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,18 +96,21 @@ public void SaveGLTFandBin(string path, string fileName)
 			gltfFile.Close();
 			binFile.Close();
 #endif
-
+			GL.sRGBWrite = true;
```
stevenvergenz: Good find! Though really, these two lines should be the only content of this PR. None of the other changes should be included, please strip them.

ziedbha: Thanks! I wanted to make a PR ASAP to people can look it at it, but I am planning on making a cleaner one as soon as possible, with only the requested changes included. Expect one within this week :)

ziedbha:
Now my fork shouldn't have conflicts with the base repo. I only edited the Exporter script. Please try exporting some models and see if the color space problem is still there.

FYI, these are the texture types that are in sRGB/Gamma space:
Emissive, BaseCoolor, Occlusion
Please let me know if that's not correct.

ziedbha:
@justinctlam I also liked the idea you had in the PR. My fix here only addresses Color space issues. So the whole idea of the ImageWrapper doesn't need to be there, as long as we can specify what textures are in what color space, we should be fine

justinctlam:
@ziedbha Sounds good : )

ziedbha:
@justinctlam checkout the newest version of this PR. I integrated changes in commit https://github.com/KhronosGroup/UnityGLTF/pull/134/commits/d72c74aa59956e61c2ffc40764d1614cf9ae0d4f

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -4,17 +4,38 @@
...
+using UnityEditor;
 
 namespace UnityGLTF
 {
 	public class GLTFSceneExporter
 	{
+        enum TextureMapType
+        {
+			//sRGB
+            Main,
```
bghgary: You've got some odd spacing issues in your changes. I think this project uses tabs.

UnityGLTF/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 5.6.5f1
+m_EditorVersion: 5.6.1f1
```
bghgary: You should probably revert this. It's going back in version?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -4,17 +4,38 @@
 using GLTF.Schema;
 using UnityEngine;
 using UnityGLTF.Extensions;
+using UnityEditor;
 
 namespace UnityGLTF
 {
 	public class GLTFSceneExporter
 	{
+        enum TextureMapType
```
bghgary: Why not add a Boolean that indicates whether the image is sRGB or not instead of a type and then a switch to determine it?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -96,21 +117,59 @@ public void SaveGLTFandBin(string path, string fileName)
...
-			{
-				Debug.Log(image.name);
-				var renderTexture = RenderTexture.GetTemporary(image.width, image.height);
-				Graphics.Blit(image, renderTexture);
-				RenderTexture.active = renderTexture;
-				var exportTexture = new Texture2D(image.width, image.height);
-				exportTexture.ReadPixels(new Rect(0, 0, renderTexture.width, renderTexture.height), 0, 0);
-				exportTexture.Apply();
-				File.WriteAllBytes(Path.Combine(path, image.name + ".png"), exportTexture.EncodeToPNG());
+        private void exportImages(string outputPath)
```
bghgary: minor coding convention issue: functions are all supposed to be upper camel case.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -529,43 +604,41 @@ private PbrMetallicRoughness ExportPBRMetallicRoughness(UnityEngine.Material mat
...
 			}
 
 			if (material.HasProperty("_Metallic"))
 			{
-				pbr.MetallicFactor = material.GetFloat("_Metallic");
+                var metallicGlossMap = material.GetTexture("_MetallicGlossMap");
+                pbr.MetallicFactor = (metallicGlossMap != null) ? 1.0 : material.GetFloat("_Metallic");
 			}
 
-			if (material.HasProperty("_Roughness"))
```
bghgary: newer versions of Unity uses _Roughness instead. I think you will break that code if you make this change.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
ziedbha: That's weird, I'll fix that!

UnityGLTF/ProjectSettings/ProjectVersion.txt
ziedbha: This was back in the first PR I made, which wasn't organized at all. My bad. There is probably more stuff to revert (only a couple of scripts should be edited...)

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
ziedbha: This was part of @justinctlam's changes. I just didn't want to add more to the code, but I can definitely do that. Thanks for pointing that out!

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
ziedbha: Got it, thanks!

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
ziedbha: Not sure what happened there, will fix it.

ziedbha: Update: since I integrated PR #133 with this one, it pulled changes in that PR too. I think @justinctlam might have some explanation as to why he took out the _Roughness factor

justinctlam: Which Unity uses _Roughness? I am on Unity 2017.3.1f1. I looked at the Standard.shader and StandardRoughness.shader and I don't see the property. Can you point me to it?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
justinctlam: I don't follow how my changes are related to sRGB?

ziedbha: They do not, which is why I added it. We need to export textures in the proper color space, or they won't look right.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
bghgary: You are right. This is my mistake. So, this never worked before then, right?

ziedbha:
I tried exporting with this PR that includes @justinctlam changes and my changes, and these are the results I am getting. I attached an exported BoomBox, Lantern, and Corset, which are 1-1 with the original models.


[boombox.zip](https://github.com/KhronosGroup/UnityGLTF/files/1791454/boombox.zip)
[corset.zip](https://github.com/KhronosGroup/UnityGLTF/files/1791457/corset.zip)
[lantern.zip](https://github.com/KhronosGroup/UnityGLTF/files/1791460/lantern.zip)

ziedbha:
For some reason, removing the Apply() line after the ReadPixels() line fixed it. After removing that line, I noticed that the GL.sRGBWrite line didn't matter anymore. I'm still trying to understand why, but this version gave me the correct results.

EDIT: It might be because of the updateMipmaps argument, which defaults to true. https://docs.unity3d.com/ScriptReference/Texture2D.Apply.html

ziedbha:
Put this on hold until Monday, I found a bug I need to fix!

blgrossMS:
Let me know where you are ready for this to get checked in. Also how your work relate to the work that is being done in PR #133 ?

ziedbha:
@blgrossMS will ping you when it's ready! The changes are simply to use the Blit method we have right now to export the images out. PR #133 did not use that method (and needed many passes on the pixels), so it was slower.

ziedbha:
@blgrossMS you can ignore this PR. However, please check PR #152

PR #137, opened 2018-03-02
3D Viewer Beta, HoloLens, Microsoft

AsoboStudio:
- Adding support for skinning
- Adding support for animations
- Fixing shader (normal maps)
- Upgrading to Unity 2017.3.1p1
- Removed some warnings
- Using Low Poly Fox to test skinning and animation repo gltf-sample-models, other pull request

### DIFF
--- a/GLTFSerialization/GLTFSerialization/Extensions/GLTFJsonExtensions.cs
+++ b/GLTFSerialization/GLTFSerialization/Extensions/GLTFJsonExtensions.cs
@@ -43,6 +43,24 @@ namespace GLTF.Extensions
 			return list;
 		}
 
+		public static List<int> ReadInt32List(this JsonReader reader)
+		{
+			if (reader.Read() && reader.TokenType != JsonToken.StartArray)
+			{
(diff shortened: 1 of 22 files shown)


blgrossMS:
Thanks for the contribution. I recently updated the repo, and there was another animation PR that came in as well, but I'll work within the next week to get these changes in as well.

blgrossMS:
As an update on this, other PR's have been adding animation fixes that got integrated that I believe took a bit from this PR. The change that I think we would like to consider from this is the error handling mechanism. @stevenvergenz can you take a look at the shader changes and see if we want to bring over?

AsoboStudio:
Thanks for the update. 
The error handling mechanism allows the app to never crash when opening corrupted or non supported GLTF files, and the app can give user explicit error messages.
The changes in shader improves very much the rendering of normal map materials; maybe you are now using other shaders.
We are happy if any part of this PR is useful to you.

blgrossMS:
Closing due to animations already being added. Created issue #204 for the error handling

Issue #135, opened 2018-03-01
repackaging GLTFSerialization as NetStandard package


vpenades:
The GLTFSerialization project is a great implementation of GLTF, from which many more developers could benefit outside the Unity environment.

I did create a small csproj targeting NetStandard 2 and the whole code of GLTFSerialization compiles without issues. So it can be trivial to add a new project to compile GLTFSerialization as a nuget package.

I can try to fork and PR it if there's an agreement on this feature.

@bghgary

blgrossMS:
This PR has been merged so I'll close the issue. Thanks again for the contribution.

PR #139, opened 2018-03-04
Added a new project to compile GLTFSerialization for NetStandard 1.3 …

vpenades:
…and generate the associated nuget package.

Also, changed a MemoryStream method from GetBuffer() to ToArray();

The ToArray(); method does the same as GetBuffer() but it's available in more platforms.

@bghgary this is related to issue #135

### DIFF
--- a/GLTFSerialization/GLTFSerialization/GLTFHelpers.cs
+++ b/GLTFSerialization/GLTFSerialization/GLTFHelpers.cs
@@ -184,7 +184,7 @@ namespace GLTF
 			{
 				using (var memoryStream = attributeAccessor.Stream as System.IO.MemoryStream)
 				{
-					bufferViewCache = memoryStream.GetBuffer();
+					bufferViewCache = memoryStream.ToArray();
 					return totalOffset;
 				}
 			}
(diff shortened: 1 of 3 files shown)


GLTFSerialization/GLTFSerialization/GLTFHelpers.cs
```diff
@@ -184,7 +184,7 @@ private static long LoadBufferView(AttributeAccessor attributeAccessor, out byte
 			{
 				using (var memoryStream = attributeAccessor.Stream as System.IO.MemoryStream)
 				{
-					bufferViewCache = memoryStream.GetBuffer();
+					bufferViewCache = memoryStream.ToArray();
```
sbtron: Sorry not sure what this is supposed to help with? Why to switch to ToArray instead of GetBuffer?

sbtron: @blgrossMS any idea for why the conditional compile for not .NETCORE? The code after it seems to be doing the same thing.

vpenades:
@sbtron

Net Standard 1.3 doesn't support GetBuffer, at first I wanted to fix it with an extension, then I fixed the code directly. Footballers is not needed.

Also, since net.core fully supports net standard, there's no need to target net core. It's already supported

blgrossMS:
GetBuffer should get the internally stored buffer that the MemoryStream is wrapping, while ToArray is going to create a Copy which we probably do not want. I'll take a look.

As another note, I recommend not merging directly from a fork's master into a repo, but creating a branch and always merging that in.

PR #144, opened 2018-03-08
Better memory management and explicit mapping between imported GLTF MinFilter and Unity FilterMode.

rawnsley:
_[Edit: Apologies for the overlapping changes in this pull request. I assumed it would be a snapshot of the forks master branch rather than a live reference. I will use explicit branches for each PR in future.]_

Improved disposal of temporary streams during import, particularly for non-binary files where multiple MemoryStreams are created but never explicitly disposed. Repeated loading large files, [such as this one](http://data.avncloud.com/activities/508020/files/public_test_scan_apollo_10_capsule.zip?date=1516446680&size=15816376), causes my Unity Android app to crash very quickly.

A closer match for Mipmap variants of nearest point sampling. Although MinFilterMode.Nearest is supported, the MipMap versions are not. [Here is an example](http://avn.cloud/3aRqqs) that uses NearestMipmapNearest.


stevenvergenz:
Thanks for your contribution! Content looks good, but could you match the project white space consensus? There are a lot of otherwise unchanged code segments that were reindented.

blgrossMS:
The IDisposable changes are great and the stream protection. The way that we handle loading data recently changed, so I'll make sure to take a look to make sure that we handle stream lifecycle correctly there. 

I'll work on getting your changes checked in.

rawnsley:
@stevenvergenz Sorry about the whitespace. I thought I had caught all the style problems before check in and I'll be a bit more diligent next time.

@blgrossMS Thanks for your help. I'm being a bit fussy about applying the pattern in this instance as the garbage collector will deal will all of this stuff eventually, but I've been chasing a memory leak and this was the first thing I thought to check.

blgrossMS:
Thanks for the PR, sorry it took so long to integrate

PR #152, opened 2018-03-29
Added texture color space fix and Pbr shader case + updated Readme

ziedbha:
You can ignore PR #134 since I decided it would be much cleaner to open a new one with a clean commit. This one contains the needed fixes to allow for proper color space conversion when exporting textures
I also added the case where the PBR shader is used (since we somehow forgot to add it when we created the new PbrMetallicRoughness shader.

In addition to this, I am working on a short documentation (Getting Started) for people who wish to help but might find the project structure complicated at first.


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -97,7 +97,8 @@ public void SaveGLTFandBin(string path, string fileName)
 			binFile.Close();
 #endif
 
-			foreach (var image in _images)
+            GL.sRGBWrite = true;
+            foreach (var image in _images)
```
blgrossMS: nit- tabs vs spaces

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -97,7 +97,8 @@ public void SaveGLTFandBin(string path, string fileName)
 			binFile.Close();
 #endif
 
-			foreach (var image in _images)
+            GL.sRGBWrite = true;
```
blgrossMS: This is only true for the base color and emmissive textures

ziedbha: @blgrossMS the issue is not as simple as sticking to the gltf specs. I played around with what you suggested a while back, and it turns out that this flag needs to be enabled for all textures. Here is the explanation that made sense to me:

Based on the Unity documentation, the default shader color space follows the PlayerSettings (under project settings). When the colorSpace is set to Linear (like we have it right now), RenderTextures will set ReadWrite space to sRGB (see https://docs.unity3d.com/ScriptReference/RenderTextureReadWrite.Default.html)

Given this, then all RenderTextures will do the following by default:
"When Linear color space is used, then by default non-HDR render textures are considered to contain sRGB data (i.e. "regular colors"), and fragment shaders are considered to output linear color values. So by default the fragment shader color value is converted into sRGB when rendering into a texture; and when sampling the texture in the shader the sRGB colors are converted into linear values. This is the sRGB read-write mode; and the Default mode matches that when linear color space is used. When this mode is set on a render texture, RenderTexture.sRGB will return true."

So when sampling, we get *linear values*. But when we render into a texture, we get *sRGB values*. Right now, we are sampling using the shader and converting into linear values for all textures.

Going back to this sRGBWrite flag, we have that: 
"This property is only relevant when Linear Color Space rendering is used. Typically when linear color space is used, non-HDR render textures are treated as sRGB data (i.e. "regular colors"), and fragment shaders outputs are treated as linear color values. So by default the fragment shader color value is converted into sRGB."

So we want that flag to always be true in order to reverse the conversion when rendering, right? I think what GL.Blit is doing is wrong at the moment because it's not reversing all the conversions we did by passing the materials through the shaders.

This is the conclusion @bghgary and I reached when we met 2 weeks ago.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
ziedbha: @blgrossMS: I don't know why VS keeps doing this, I thought I fixed my default spacing settings. What are the default settings I should set? Thanks for catching this!

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
blgrossMS: That sounds right. If the exported images look correct then this should be good.

blgrossMS:
@ziedbha Once you resolve the conflits I'll merge it in

ziedbha:
No need for this since @justinctlam fixed it in the newly added PR!

PR #155, opened 2018-04-04
Add support for binary GLB Export

RPSpicer:
This changeset adds the ability to export binary GLBs. I'm looking for feedback to iterate and improve!

What's working:
-JSON and Meshes.
-Passes the GLTF validator with what I've thrown at it.

What's not working:
-Current implementation assumes little-endian. Need to check for big-endian and swap endianness if required, or at least throw a meaningful exception.
-Doesn't currently support embedding images in the binary chunk.

What needs improvement:
-Allocating separate MemoryStreams for bin and json chunks is inefficient. If I could calculate the size of the bin chunk before I started serializing the JSON chunk, I could write everything directly into a single Stream.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
@@ -122,6 +122,11 @@ namespace UnityGLTF
 			return _root;
 		}
 
+		/// <summary>
+		/// Writes a binary GLB file with filename at path.
+		/// </summary>
+		/// <param name="path">File path for saving the binary file</param>
+		/// <param name="fileName">The name of the GLTF file</param>
 		public void SaveGLB(string path, string fileName)
 		{
 			Stream binStream = new MemoryStream();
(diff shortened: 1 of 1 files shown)


blgrossMS:
I have a branch that is taking an attempt to solve this problem more holistically: https://github.com/blgrossMS/UnityGLTF/tree/msft_bug_fixes_and_glb_update

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -115,6 +122,88 @@ public GLTFSceneExporter(Transform[] rootTransforms, RetrieveTexturePathDelegate
...
+			int length;
+			while ((length = input.Read(buffer, 0, buffer.Length)) > 0)
+			{
+				output.Write(buffer, 0, length);
+			}
+		}
+
+		private void AlignToBoundary(Stream stream, byte pad = (byte)' ', int boundary = 4)
+		{
+			int offset = (int)(stream.Length % boundary);
```
blgrossMS: this will bloat by 4 bytes on the zero case. Better version is: 

```
public static uint CalculateAlignment(uint currentSize, uint byteAlignment)
		{
			return (currentSize + byteAlignment - 1) / byteAlignment * byteAlignment;
		}
```

blgrossMS:
Take care of my comment and I'll merge it in

RPSpicer:
Thank you for feedback; updated with your code.  Looking forward to the holistic rework, which seems like the correct long-term approach. Lots of improvements in there!

Issue #150, opened 2018-03-20
[glTF Validator] Accessor's total byteOffset 1126394 isn't a multiple of componentType length 4.


sonygod:
[glTF Validator] Accessor's total byteOffset 1126394 isn't a multiple of componentType length 4.
[glTF Validator] Accessor's total byteOffset 1135226 isn't a multiple of componentType length 4.
[glTF Validator] Accessor's total byteOffset 1144058 isn't a multiple of componentType length 4.
[glTF Validator] Accessor's total byteOffset 1155834 isn't a multiple of componentType length 4.
[glTF Validator] Accessor's total byteOffset 1169438 isn't a multiple of componentType length 4.
[glTF Validator] Accessor's total byteOffset 1174526 isn't a multiple of componentType length 4.
[glTF Validator] Accessor's total byteOffset 1179614 isn't a multiple of componentType length 4.
[glTF Validator] Accessor's total byteOffset 1186398 isn't a multiple of componentType length 4.
glTF Property

The root object for a glTF asset.

![image](https://user-images.githubusercontent.com/748320/37646002-6a914eb4-2c63-11e8-9e31-714e6c619a34.png)
![image](https://user-images.githubusercontent.com/748320/37646019-7319a360-2c63-11e8-9974-d39bc5736bdc.png)

sonygod:
[14.zip](https://github.com/KhronosGroup/UnityGLTF/files/1828721/14.zip)

sonygod:
[testmodel2.zip](https://github.com/KhronosGroup/UnityGLTF/files/1832013/testmodel2.zip)
unity 2017.3.x

ziedbha:
I'll look into this! This is I bug I also found :)

Neuxxx:
[GLTFSceneExporter.zip](https://github.com/KhronosGroup/UnityGLTF/files/2026314/GLTFSceneExporter.zip)

I have run into the same Bug. In my case the Mixed-Reality-Viewer from Windows don't open my exported GLB Files.
I have add a little Methode to the exporter 
`private long AppendToBufferMultiplyOf4(long byteOffset, long byteLength)
		{
			var moduloOffset = byteLength % 4;
			if (moduloOffset > 0)
			{
				for (int i = 0; i < (4 - moduloOffset); i++)
				{
					_bufferWriter.Write((byte)0x00);
				}
				byteLength = _bufferWriter.BaseStream.Position - byteOffset;
			}

			return byteLength;
		}`

[...]

PR #160, opened 2018-04-06
Added animation export (runtime AnimatorController)

ziedbha:
@bghgary worked on this previously, and I think his implementation was correct, so I went ahead and mapped his work to this project. Most of what I did is very similar to what Gary did!

To explain the logic: the newest Unity builds use an Animator object instead of the legacy Animation object. An Animator holds an AnimatorController, which holds a DFA of animation states, and each state holds an animation clip. We just need to export the clips that animate properties (position, rotation, scale, and morph targets). 

Since there was a lot of code to write, I had to extend the base Exporter script. I also extended it so we can export different types of Accessors, just like Gary did in his project).

I am not able to test this properly since I don't have nice animated assets, so maybe we should add more assets that use the Animator component. Once this is done I can review this PR properly!

All comments are appreciated :)


UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs
```diff
@@ -0,0 +1,219 @@
...
+			};
+
+			_root.Accessors.Add(accessor);
+
+			return id;
+		}
+
+		private static int Align(int value, int size)
+		{
+			var remainder = value % size;
```
ziedbha: I think this is the fix needed to fix many of the Accessor bugs we have (see issue #150)

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+				keyIndex => weightCurves.Select(curve => curve.keys[keyIndex].value / 100),
+				keyIndex => weightCurves.Select(curve => curve.keys[keyIndex].outTangent / 100),
+				time => weightCurves.Select(curve => curve.Evaluate(time) / 100),
+				values => this.ExportData(values.SelectMany(value => value)));
+		}
+
+		private AnimationSampler ExportAnimationSampler<T>(IEnumerable<AnimationCurve> curves, Func<int, T> getInTangent, Func<int, T> getValue, Func<int, T> getOutTangent, Func<float, T> evaluate, Func<IEnumerable<T>, AccessorId> exportData)
+		{
+			//decide if we should include baking animations or not
+			if (/*bakeAnimations is False &&*/ CanExportCurvesAsSpline(curves))
```
bghgary: This baking code can use a bit of work. We should at least not write the same output value if they are close enough. I also just picked an arbitrary 30 fps to bake the animation which might or might not be right.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs
```diff
@@ -0,0 +1,219 @@
...
+
+		private AccessorId ExportAccessor(BufferViewId bufferViewId, GLTFComponentType componentType, int count, GLTFAccessorAttributeType type, IEnumerable<object> min, IEnumerable<object> max, string name = null)
+		{
+			Accessor accessor = new Accessor
+			{
+				BufferView = bufferViewId,
+				ComponentType = componentType,
+				Count = count,
+				Type = type,
+				Min = (List<double>) min,
```
bghgary: This doesn't work for me. It throws a runtime exception.

bghgary: InvalidCastException: Cannot cast from source type to destination type.
UnityGLTF.GLTFSceneExporter.ExportAccessor (GLTF.Schema.BufferViewId bufferViewId, GLTFComponentType componentType, Int32 count, GLTFAccessorAttributeType type, IEnumerable`1 min, IEnumerable`1 max, System.String name) (at Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs:40)
UnityGLTF.GLTFSceneExporter.ExportData (GLTFAccessorAttributeType type, GLTFComponentType componentType, Int32 componentSize, Int32 count, IEnumerable`1 min, IEnumerable`1 max, Int32 byteLength, System.Action`1 writeData, System.String accessorName) (at Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs:25)

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
+﻿using System;
+using System.Collections.Generic;
+using GLTF.Schema;
+using UnityEngine;
+using UnityEditor.Animations;
+using UnityEditor;
```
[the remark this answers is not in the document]
blgrossMS: That's right

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/BinaryWriterExtensions.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System.IO;
+using UnityEngine;
+using VectorTypes;
+
+namespace VectorTypes
+{
+	//New vector types
+	public struct ByteVector4
```
blgrossMS: all of this logic should be in the GLTFSerilization layer

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+using System.Linq;
+
+namespace UnityGLTF
+{
+	public partial class GLTFSceneExporter
+	{
+		private enum Property
+		{
+			m_LocalPosition = GLTFAnimationChannelPath.translation,
```
blgrossMS: m_ is a more of a C++ paradigm. Just use _localPosition;

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+			m_LocalRotation = GLTFAnimationChannelPath.rotation,
+			m_LocalScale = GLTFAnimationChannelPath.scale,
+			blendShape = GLTFAnimationChannelPath.weights,
+		}
+
+		//holds already exported animation clips
+		private readonly Dictionary<object, int> _animationCache = new Dictionary<object, int>();
```
blgrossMS: Replace this with the asset cache and don't use object

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+					var animatorController = (AnimatorController)animator.runtimeAnimatorController;
+					if (animatorController == null)
+					{
+						//no animations attached
+						continue;
+					}
+
+					foreach (var layer in animatorController.layers)
+					{
+						var animationStates = layer.stateMachine.states;
```
blgrossMS: I think using var throughout this make it a lot harder to read

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+							if (state.motion is BlendTree)
+							{
+								throw new NotSupportedException();
+							}
+
+							var animationClip = (AnimationClip)state.motion;
+							if (animationClip != null)
+							{
+								//export animation clip from each state of the state machine
+								this.ExportAnimation(gameObject, animationClip);
```
blgrossMS: this. isn't needed

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+						}
+					}
+				}
+			}
+		}
+
+		private void ExportAnimation(GameObject gameObject, AnimationClip unityAnimationClip)
+		{
+			int index;
+			if (_animationCache.TryGetValue(unityAnimationClip, out index))
```
blgrossMS: can be simplified to out int index

blgrossMS: this should match asset cache

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+				Name = unityAnimationClip.name,
+				Channels = channels,
+				Samplers = samplers
+			};
+			
+			_root.Animations.Add(animation);
+		}
+
+		//for all editor curve bindings of an animation clip, return a map: path -> (property -> member)
+		private Dictionary<string, Dictionary<string, Dictionary<string, EditorCurveBinding>>> GroupAnimationCurveBindings(IEnumerable<EditorCurveBinding> editorCurveBindings)
```
blgrossMS: GetGroupAnimationCurveBindings?

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+			{
+				//already exported this clip, skip
+				return;
+			}
+
+			var channels = new List<AnimationChannel>();
+			var samplers = new List<AnimationSampler>();
+
+			//for each: path -> (property -> member)
+			foreach (var pairPath in this.GroupAnimationCurveBindings(AnimationUtility.GetCurveBindings(unityAnimationClip)))
```
blgrossMS: break this out to make it more readable

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+			{
+				var path = pairPath.Key;
+				var propertyCurves = pairPath.Value;
+
+				//for each: property -> member
+				foreach (var pairProperty in propertyCurves)
+				{
+					var property = (Property)Enum.Parse(typeof(Property), pairProperty.Key);
```
blgrossMS: add a comment for what this is doing

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+				memberBindings.Add(member, editorCurveBinding);
+			}
+
+			return bindings;
+		}
+
+		private AnimationSampler ExportAnimationSamplerPosition(AnimationCurve curveX, AnimationCurve curveY, AnimationCurve curveZ)
+		{
+			return this.ExportAnimationSampler(
+				new[] { curveX, curveY, curveZ },
```
blgrossMS: explciity declare the type of array

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+					return false;
+				}
+			}
+
+			return true;
+		}
+
+		private static Vector3 GetRightHandedVector(Vector3 value)
+		{
+			return new Vector3(value.x, value.y, -value.z);
```
blgrossMS: are you doing vector conversion here? why? Is it to convert between spaces? We have classes already that should do this conversion

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+		}
+
+		private static Vector4 GetRightHandedVector(Vector4 value)
```
blgrossMS: these should already be implemented in SchemaExtensions

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
blgrossMS: comment?

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+			return this.ExportAnimationSampler(
+				new[] { curveX, curveY, curveZ },
+				keyIndex => GetRightHandedVector(new Vector3(curveX.keys[keyIndex].inTangent, curveY.keys[keyIndex].inTangent, curveZ.keys[keyIndex].inTangent)),
+				keyIndex => GetRightHandedVector(new Vector3(curveX.keys[keyIndex].value, curveY.keys[keyIndex].value, curveZ.keys[keyIndex].value)),
+				keyIndex => GetRightHandedVector(new Vector3(curveX.keys[keyIndex].outTangent, curveY.keys[keyIndex].outTangent, curveZ.keys[keyIndex].outTangent)),
+				time => GetRightHandedVector(new Vector3(curveX.Evaluate(time), curveY.Evaluate(time), curveZ.Evaluate(time))),
+				values => this.ExportData(values));
+		}
+
+		private static Quaternion CreateNormalizedQuaternion(float x, float y, float z, float w)
```
blgrossMS: this should already be implemented by unity

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs
```diff
@@ -0,0 +1,219 @@
...
+
+			_root.Accessors.Add(accessor);
+
+			return id;
+		}
+
+		private static int Align(int value, int size)
+		{
+			var remainder = value % size;
+			return (remainder == 0 ? value : checked(value + size - remainder));
```
blgrossMS: use (value + size - 1) / size * size instead. It will do the alignment without the branch

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs
```diff
@@ -0,0 +1,219 @@
...
+		{
+			return this.ExportData(
+				GLTFAccessorAttributeType.SCALAR,
+				GLTFComponentType.Float,
+				sizeof(float),
+				values.Count(),
+				minMax ? new object[] { values.Min() } : null,
+				minMax ? new object[] { values.Max() } : null,
+				sizeof(float) * values.Count(),
+				binaryWriter => values.ForEach(value => binaryWriter.Write(value)));
```
blgrossMS: Shouldn't the binary write take an array of values?

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
blgrossMS: Also weird for enums. Do:
LocalPosition,
LocalRotation,
LocalScale,
BlendShape

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+			{
+				var firstCurve = curves.First();
+
+				var input = new float[firstCurve.keys.Length];
+				var output = new T[firstCurve.keys.Length * 3];
+				for (int keyIndex = 0; keyIndex < firstCurve.keys.Length; keyIndex++)
+				{
+					input[keyIndex] = firstCurve.keys[keyIndex].time;
+
+					output[keyIndex * 3 + 0] = getInTangent(keyIndex);
```
blgrossMS: explain the magic numbers

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+					Interpolation = GLTF.Schema.InterpolationType.CUBICSPLINE,
+					Output = exportData(output),
+				};
+			}
+			else
+			{
+				var input = new List<float>();
+				var output = new List<T>();
+				var maxTime = curves.Max(curve => curve.keys.Last().time);
+				for (float time = 0; time < maxTime; time += 1.0f / 30.0f)
```
blgrossMS: this should be a constant. Why 30 ms?

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs
```diff
@@ -0,0 +1,219 @@
...
+			return id;
+		}
+
+		private static int Align(int value, int size)
+		{
+			var remainder = value % size;
+			return (remainder == 0 ? value : checked(value + size - remainder));
+		}
+
+		private AccessorId ExportData(IEnumerable<ushort> values)
```
blgrossMS: Anything that has to do with writing data to accessors should live in the GLTFSerialization layer

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
bghgary: We should put up a coding convention document so contributors know what to use.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
bghgary: I mentioned this above. It's arbitrarily set at 30fps right now. Maybe this should be a setting.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
ziedbha: Will change it all

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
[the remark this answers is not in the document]
ziedbha: One way to do this is to add an editor script that parses through the animation curves in editor mode, stores them somewhere as an exposed variable, then in runtime mode get that information. This is redundant, but would at least allow for "runtime" export.. thoughts?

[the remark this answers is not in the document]
bghgary: I don't have a good solution for this. Maybe we need to ask Unity on this one.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
ziedbha: from what I've seen, Unity expects the quaternions to be normalized and does not provide a method to check/do this.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs
ziedbha: in the scene exporter we have right now, we are writing component by component, vector by vector. We can extend it to take in an array though.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
bghgary: This is as the function name says, to convert the left handed data to right handed which is a space conversion. If there already is one, we should use it.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,335 @@
...
+		//holds already exported animation clips
+		private readonly Dictionary<object, int> _animationCache = new Dictionary<object, int>();
+
+		private void ExportAnimations(IEnumerable<Transform> transforms)
+		{
+			foreach (var transform in transforms)
+			{
+				//use Animator instead of Animation for support of newest features
+				var gameObject = transform.gameObject;
+				foreach (var animator in gameObject.GetComponentsInChildren<Animator>())
```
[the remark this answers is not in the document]
ziedbha: Can work on them once this version is done

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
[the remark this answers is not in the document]
ziedbha: How can we do that?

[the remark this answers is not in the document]
bghgary: We can try the forums or I can reach out.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Data.cs
ziedbha: Does that apply when ExportData calls ExportAccessor & ExportBufferView, which both rely on the root GLTF object being constructed, and the fact that we are only building to one Buffer?

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
ziedbha: I understand what it does, but SchemaExtensions does not seem to have a function that does this exact space conversion. Am I missing something?

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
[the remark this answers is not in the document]
ziedbha: I checked some recent threads and most people seem to repeat the fact that this not currently feasible. It might be better to reach out to someone from Unity directly, if possible.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
bghgary: I haven't looked the SchemaExrensions, so I can't help with that one. @blgrossMS?

ziedbha:
@bghgary @blgrossMS I pushed a commit in which I:

- **Fixed code from a style standpoint**, as demanded. I commented the complicated functions as needed
- **Refactored the entire Data export functions into the serializer under DataExporter.cs**. We now have to specify which buffer writer we are using the serialize, as well as which buffer to write to. I know that in the project we are exporting to one single buffer (buffer 0), but I thought that since the Serializer is project agnostic, we should be able to specify what buffer to export to
- **Extended the BinaryWriter under the serializer**. This led to the need to convert all Unity Math components to GLTF Math components using the SchemaExtensions. For example, we now won't write the Unity Vector3 immediately, but we would convert it to a GLTF Vec3 then write it. This is inefficient: I couldn't figure out a way to avoid the conversion while extending the binary writer in the serializer.
- **Fixed some bugs** such as the cast to double

I DID NOT:
- **Improve the baking code**. I need someone to explain to me how we want to approach this since I do not have experience with baking. I have some high level ideas on how it works though
- **Improve the caching**. I will get to this soon, I just wanted to push something so that we are able to review step by step

I also exported the animated RiggedFigure, and I got this output:
[RiggedFigureExported.zip](https://github.com/KhronosGroup/UnityGLTF/files/2019961/RiggedFigureExported.zip)

Not sure how to test this properly, but it seems like we improved on this a bit.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Animation.cs
```diff
@@ -0,0 +1,360 @@
...
+		{
+			return new Vector3(value.x, value.y, -value.z);
+		}
+
+		private static Vector4 GetRightHandedVector(Vector4 value)
+		{
+			return new Vector4(value.x, value.y, -value.z, -value.w);
+		}
+
+		private static Quaternion GetRightHandedQuaternion(Quaternion value)
```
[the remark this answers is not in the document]
ziedbha: Should I change that? And do you also want me to add the GetRightHandedVector in the SchemaExtensions?

blgrossMS:
Closing for now due to lack of progress. Should be built on top of later for animation work.

PR #164, opened 2018-04-11
Added GLTF EditorWindow and settings for exporter

JasperDre:
Hi,

It would be nice to be able to customize certain behavior of the tool. An example is the export of textures/images. Recently this has been programmed to export using the full file path. However, depending on the pipeline it might not be as useful to do so (like in my case). To make it easier to customize the tool I added a settings tab, which contains some of the booleans and useful information. Will this be a nice contribution?

settings


blgrossMS:
Thank your for the contribution!

PR #168, opened 2018-04-12
Jcopic/indexfix

JohnCopicMS:
Fix for models that have >64k verticies per mesh. This fix only works in Unity 2017.3 or greater, where int-sized vertex lists are supported.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -17,815 +17,815 @@ using Debug = UnityEngine.Debug;
 
 namespace UnityGLTF
 {
-    public struct MeshConstructionData
-    {
-        public MeshPrimitive Primitive { get; set; }
-        public Dictionary<string, AttributeAccessor> MeshAttributes { get; set; }
-    }
-
-    public class GLTFSceneImporter
-    {
-        public enum ColliderType
-        {
-            None,
-            Box,
-            Mesh
-        }
-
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1115,15 +1115,18 @@ protected IEnumerator ConstructUnityMesh(MeshConstructionData meshConstructionDa
...
-                    ? meshAttributes[SemanticProperties.POSITION].AccessorContent.AsVertices.ToUnityVector3Raw()
-                    : null,
-                normals = primitive.Attributes.ContainsKey(SemanticProperties.NORMAL)
-                    ? meshAttributes[SemanticProperties.NORMAL].AccessorContent.AsNormals.ToUnityVector3Raw()
-                    : null,
+			// todo optimize: There are multiple copies being performed to turn the buffer data into mesh data. Look into reducing them
+			UnityEngine.Mesh mesh = new UnityEngine.Mesh
+			{
+#if UNITY_2017_3_OR_NEWER
+				indexFormat = vertexCount > 65535 ? IndexFormat.UInt32 : IndexFormat.UInt16,
```
blgrossMS: Can you have a case where you have less than 65535 indices that are being stored as 32 bit integers?

JohnCopicMS: Even if they were stored in 32 bit values you wouldn't lose anything by putting them in to shorts in the mesh.

PR #172, opened 2018-04-18
Adding skins exporting feature

ziedbha:
Much like what @bghgary did in his personal repo, I extended the exporter script to support skin exporting. 
In PR #160, I added an extension to support exporting accessors/bufferviews for many data types (vectors, matrices...). To export the bind matrices, I use that function.

It's important to review PR #160 before this one.


UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
```diff
@@ -0,0 +1,106 @@
...
+			{
+				BindPoses = skinnedMeshRenderer.sharedMesh.bindposes,
+				RootBone = skinnedMeshRenderer.rootBone,
+				Bones = skinnedMeshRenderer.bones,
+			};
+
+			SkinId skinId;
+			if (_skinCache.TryGetValue(skin, out skinId))
+			{
+				//already exported this clip, skip
```
bghgary: s/clip/skin

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
```diff
@@ -0,0 +1,106 @@
...
+			{
+				Id = _root.Skins.Count,
+				Root = this._root
+			};
+
+			_root.Skins.Add(new GLTF.Schema.Skin
+			{
+				//references ExportData in Animation PR
+				InverseBindMatrices = this.ExportData(skin.BindPoses.Select(bindpose => GetRightHandedMatrix(bindpose))),
+				Skeleton = this._nodeCache[skinnedMeshRenderer.transform],
```
bghgary: There is no need to write out Skeleton. See https://github.com/KhronosGroup/glTF/pull/1195.

bghgary: Also, if the parent of the skin root node has scale, this code doesn't work. Unity ignores scale above the skinnedMeshRenderer, but glTF does not.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
```diff
@@ -0,0 +1,106 @@
...
+				Joints = skin.Bones.Select(bone => this._nodeCache[bone.transform]).ToList(),
+			});
+
+			return skinId;
+		}
+
+		//Skin struct for caching purposes
```
blgrossMS: shouldn't all of this information already be cconstructed as a GLTF.Schema object?

ziedbha: There isn't a 1-1 mapping from the Unity definition of the skin to the GLTF one. We need someway to compare unity skins so I can add them to the cache (since Unity skins are not defined under one struct)

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
ziedbha: In that case should I grab the localScale of the parent and multiply it by the child's localScale?

bghgary: I'm not sure what the correct math is yet. For now, you can probably get away with a warning.

bghgary: Sorry, what I meant by warning is to detect that there is a parent scale and call [Debug.LogWarning](https://docs.unity3d.com/ScriptReference/Debug.LogWarning.html)

ziedbha: Sorry - I did not not know that was the standard protocol.
Fixed!

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
```diff
@@ -0,0 +1,111 @@
...
+			}
+
+			skinId = new SkinId
+			{
+				Id = _root.Skins.Count,
+				Root = this._root
+			};
+
+			Transform parentTransform = skinnedMeshRenderer.GetComponentInParent<Transform>();
+			if (parentTransform != null && parentTransform.localScale != Vector3.one)
```
bghgary: This isn't enough. You need to check all the levels not just one level above. I also can't remember what happens if `skinnedMeshRenderer.transform` itself has scale.

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
```diff
@@ -0,0 +1,111 @@
...
+				return skinId;
+			}
+
+			skinId = new SkinId
+			{
+				Id = _root.Skins.Count,
+				Root = this._root
+			};
+
+			Transform parentTransform = skinnedMeshRenderer.GetComponentInParent<Transform>();
```
bghgary: simpler way to express this is `skinnedMeshRenderer.transform.parent`

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
ziedbha: My bad, for some reason I had the idea that each node would check its parent anyways, which is not the case at all since not all nodes necessarily have a skin. Thank you!

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
ziedbha: Got it!

UnityGLTF/Assets/UnityGLTF/Scripts/Extensions/GLTFSceneExporter.Skinning.cs
ziedbha: @bghgary I read this https://answers.unity.com/questions/147816/how-to-avoid-scaling-heritage-when-parenting.html, and it seems like the object is only affected by the direct parent's scale. Do we still need to propagate the warning to all parents (which might be costly)

bghgary: I don't think that thread has anything to do with this. You can probably get away with checking [lossyScale](https://docs.unity3d.com/ScriptReference/Transform-lossyScale.html).

blgrossMS:
Closing for now due to lack of progress. Should be built on top of later for animation work.

Issue #216, opened 2018-08-07
Loading without Blocking


cwfitzgerald:
I am attempting to write a GLTF-based world viewer with incremental loading. The assets are fairly large, so the GLTF models have to be added every time we want to load a new tile of the world. The issue we are currently having is that loading with GLTF blocks the main thread for about 500ms (loading 200K tri meshes). Unfortunately, even with the multithreading flag set to on, it doesn't help the large amount of blocking that happens. Is there any established way to deal with long periods of blocking? 

My current way for loading is to create a game object with a new GLTF component for each tile, adding and removing game objects as needed.

blgrossMS:
Hi @cwfitzgerald,

The main problem is that before 2018.1/2 all Unity operations had to happen on the main thread. This has changed a bit with their job system which should allow us to multithread more parts of our load.

Historically to account for this problem we yield between major operations today (mesh load, texture load, material load etc).

One problem with runtime loading today is that all textures have to load on main thread (which can take on the magnitude of 100's of ms depending on type of device and size of texture) since glTF supports non GPU compressed textures as part of the main specification.

Depending on your application you can pack into your assets GPU packed textures as an extension and use that to improve your loading time.

We have an issue #9 which addresses the problem we have with texture load perf. @JohnCopicMS I believe is actively looking into this problem.

Creation of the Unity mesh also happens on the main thread. If the mesh creation is taking a while, we could take a PR which breaks the mesh load into segments (by splitting the mesh). 

However, I think the best solution to the problem is going to be a refactor which takes adantage of Unity's job system to parallelize and background thread the creation of the Unity objects. 

I might also recommend that you try using a LOD system to reduce the complexity of the initial models you are loading and load in the more complex models later. Take a look at the msft_lod extension: https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Vendor/MSFT_lod 

Hope this helps! If you do end up working on this problem at all, feel free to post and ask questions 😄

Zichuan-Wang:
Hi there, I made a PR #221 that added multithreading support to the importer. Please check it out! @cwfitzgerald @blgrossMS

PR #221, opened 2018-08-23
Improving performance by adding multithreading support

Zichuan-Wang:
Hi there,

As some of you may have experienced, currently the UnityGLTF library blocks the main thread when importing models containing complex geometry and/or textures. Although the coroutine approach is used in the project, sometimes multithreading may be a better solution as it provides a **non-blocking behavior as well as better performance (in some situations)**. 

**Implementation**
I noticed a parameter `isMultithreaded` in `GLTFSceneImporter.LoadScene` that is never actually used, changed it to a public field and used it as the switch. I put some presumably time-consuming work on other threads (rather than the main thread), which includes `LoadStream`, `ParseJson`, `stream.Read`, `buildMeshAttributesThread`, and `GLTFHelpers.BuildMeshAttributes`. 

I have to change the `ILoader` interface to add support for non-coroutine based `LoadStreamSync` method, as coroutine methods cannot be easily executed on threads and doesn't really make sense for created threads anyway. I also have to change the signature of `ParseJson` in `GLTFParser.cs` to make it thread-friendly.

**Performance**
HUGE frame rate increase for certain models containing complex geometry and/or textures. I observed up to 200% fps increase for [this model](https://sketchfab.com/models/9736bfd2df1444cdab54adfc04c6a659) on a Windows desktop computer. In general, it should increase your fps and/or decrease the main thread freezing time. It should not make any tangible negative impact to the loading time, if no faster loading is observed.

**Limitations**
As mentioned in #9, `Texture2D.LoadImage` is slow and there's no obvious way around it. Same with `Texture2D.Apply`.

**Unity Job System?**
The recently introduced Unity Job System seems good, but currently it is still limited in a number of ways. Also, I don't believe it is very backward compatible.

Reference: #216

### DIFF
--- a/GLTFSerialization/GLTFSerialization/GLTFParser.cs
+++ b/GLTFSerialization/GLTFSerialization/GLTFParser.cs
@@ -19,7 +19,7 @@ namespace GLTF
 			public uint FileLength { get; set; }
 		}
 		
-		public static GLTFRoot ParseJson(Stream stream, long startPosition = 0)
+		public static void ParseJson(Stream stream, ref GLTFRoot gltfRoot, long startPosition = 0)
 		{
 			stream.Position = startPosition;
 			// Check for binary format magic bytes
@@ -32,7 +32,7 @@ namespace GLTF
 				stream.Position = startPosition;
 			}
 
-			return GLTFRoot.Deserialize(new StreamReader(stream));
+			gltfRoot = GLTFRoot.Deserialize(new StreamReader(stream));
 		}
 		
 		// Moves stream position to binary chunk location
--- a/GLTFSerialization/Tests/GLTFSerializationTests/GLTFLoaderTest.cs
+++ b/GLTFSerialization/Tests/GLTFSerializationTests/GLTFLoaderTest.cs
@@ -24,7 +24,8 @@ namespace GLTFSerializationTests
 			FileStream gltfStream = File.OpenRead(GLTF_PATH);
 
 			GLTFRoot.RegisterExtension(new TestExtensionFactory());
-			GLTFRoot gltfRoot = GLTFParser.ParseJson(gltfStream);
+			GLTFRoot gltfRoot = null;
+			GLTFParser.ParseJson(gltfStream,ref gltfRoot);
 			GLTFJsonLoadTestHelper.TestGLTF(gltfRoot);
 		}
 
@@ -33,8 +34,9 @@ namespace GLTFSerializationTests
 		{
 			Assert.IsTrue(File.Exists(GLTF_PBR_SPECGLOSS_PATH));
 			FileStream gltfStream = File.OpenRead(GLTF_PBR_SPECGLOSS_PATH);
-			
-			GLTFRoot gltfRoot = GLTFParser.ParseJson(gltfStream);
+
+			GLTFRoot gltfRoot = null;
+			GLTFParser.ParseJson(gltfStream, ref gltfRoot);
 
 			Assert.IsNotNull(gltfRoot.ExtensionsUsed);
 			Assert.IsTrue(gltfRoot.ExtensionsUsed.Contains(KHR_materials_pbrSpecularGlossinessExtensionFactory.EXTENSION_NAME));
@@ -57,7 +59,8 @@ namespace GLTFSerializationTests
 		{
 			Assert.IsTrue(File.Exists(GLB_PATH));
 			FileStream gltfStream = File.OpenRead(GLB_PATH);
-			GLTFRoot gltfRoot = GLTFParser.ParseJson(gltfStream);
+			GLTFRoot gltfRoot =null;
+			GLTFParser.ParseJson(gltfStream, ref gltfRoot);
 			GLTFJsonLoadTestHelper.TestGLB(gltfRoot);
 		}
 	}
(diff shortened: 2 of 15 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
```diff
@@ -336,14 +336,16 @@ private GameObject CreateGLTFScene(string projectFilePath)
...
                 var loader = new GLTFSceneImporter(gLTFRoot, fileLoader, stream);
 
                 loader.MaximumLod = _maximumLod;
+	            loader.isMultithreaded = true;
 
-                // HACK: Force the coroutine to run synchronously in the editor
-                var stack = new Stack<IEnumerator>();
-                stack.Push(loader.LoadScene(isMultithreaded: true));
+				// HACK: Force the coroutine to run synchronously in the editor
+				var stack = new Stack<IEnumerator>();
```
StephenHodgson: formatting seems to be off here. Make sure you're using tabs

Zichuan-Wang: Actually I'm using tabs here, but the rest of this file seems to be using spaces. Should I change the tabs to spaces?

StephenHodgson: Probably, I think the majority of the project is using tabs.

StephenHodgson: I think in this case whichever keeps the formatting changes to a minimum is best to see the diff.

Zichuan-Wang: Issue fixed!

GLTFSerialization/GLTFSerialization/GLTFParser.cs
```diff
@@ -19,7 +19,7 @@ internal struct GLBHeader
 			public uint FileLength { get; set; }
 		}
 		
-		public static GLTFRoot ParseJson(Stream stream, long startPosition = 0)
+		public static void ParseJson(Stream stream, ref GLTFRoot gltfRoot, long startPosition = 0)
```
blgrossMS: Should probably be out instead of ref

GLTFSerialization/Tests/GLTFSerializationTests/GLTFLoaderTest.cs
```diff
@@ -57,7 +59,8 @@ public void LoadGLBFromStream()
 		{
 			Assert.IsTrue(File.Exists(GLB_PATH));
 			FileStream gltfStream = File.OpenRead(GLB_PATH);
-			GLTFRoot gltfRoot = GLTFParser.ParseJson(gltfStream);
+			GLTFRoot gltfRoot =null;
```
blgrossMS: nit spacing

UnityGLTF/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
```diff
@@ -17,11 +17,14 @@ public class WebRequestLoader : ILoader
 	{
 		public Stream LoadedStream { get; private set; }
 
+		public bool HasSyncLoadMethod { get; private set; }
```
blgrossMS: Can be HasSyncLoadMethod => false;

UnityGLTF/Assets/UnityGLTF/Scripts/Loader/FileLoader.cs
```diff
@@ -15,9 +15,12 @@ public class FileLoader : ILoader
 		private string _rootDirectoryPath;
 		public Stream LoadedStream { get; private set; }
 
+		public bool HasSyncLoadMethod { get; private set; }
```
blgrossMS: HasSyncLoadMethod => true

GLTFSerialization/GLTFSerialization/GLTFParser.cs
Zichuan-Wang: I was starting to fix it but then noticed you had done it. Thanks!

UnityGLTF/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
Zichuan-Wang: I noticed you have started CI. Are you OK with leaving it as it is?

blgrossMS: Yeah it's not supported in this version of .NET

Zichuan-Wang: Got it!

Zichuan-Wang:
@blgrossMS It seems like the CI is failing due to some configuration problems... Could you maybe take a look at it?

blgrossMS:
Yup I just have to merge master, I fixed it there

Zichuan-Wang:
Cool, let me know if you need anything from me.

PR #228, opened 2018-09-06
VS solution for building DLL from UnityGLTF sources

stevenvergenz:
A Unity project is not a particularly useful delivery format for what is ultimately a library for other applications to consume. This PR adds `UnityGLTF/UnityGLTF-dll.csproj`, which describes all the same scripts as the Unity project, but packages them as a DLL (with dependencies on GLTFSerialization.dll and UnityEngine.dll). Note that this project excludes the Tests folder and anything else with UnityEditor dependencies.

### DIFF
--- a/UnityGLTF/.gitignore
+++ b/UnityGLTF/.gitignore
@@ -29,6 +29,9 @@ ExportedObj/
 *.svd
 *.pdb
 
+!*-dll.csproj
+!*-dll.sln
+
 # Unity3D generated meta files
 *.pidb.meta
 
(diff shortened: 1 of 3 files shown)


stevenvergenz:
@emackey There is lots of tuning to be done on the csproj, particularly around the Unity references. Currently it looks in a hard-coded folder for your Unity DLLs, so if you used a non-standard install location they won't be found. You can modify the project file directly in a text editor, just copy-paste another path line and modify to your needs. The generated DLL should be good for whatever version of Unity it's built against.

Issue #251, opened 2018-09-27
RoughnessFactors of materials get ignored


tolen23:
When loading a gltf for which a roughnessFactor has been specified under property pbrMetallicRoughness, this factor gets obviously ignored.

Example gltf: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/OrientationTest

The BaseCubeMesh there is associated with the following material:
```
"materials" : [
        {
            "name" : "BaseMaterial", 
            "pbrMetallicRoughness" : {
                "baseColorFactor" : [
                    0.3401939868927002, 
                    0.6781029105186462, 
                    1.0, 
                    1.0
                ], 
                "roughnessFactor" : 0.5
            }
        },
        [...]
]
```

Expected behaviour after loading this into Unity would be to see Roughness set to 0.5. Instead it is always set to 1.0:

![material_shader](https://user-images.githubusercontent.com/43646039/46156658-3d987380-c27a-11e8-845e-2d68a03cb016.PNG)

This can also be reproduced with the standard material example from the gltf 2.0 spec:

```
"materials": [
        {
            "name": "gold",
            "pbrMetallicRoughness": {
                "baseColorFactor": [ 1.000, 0.766, 0.336, 1.0 ],
                "metallicFactor": 1.0,
                "roughnessFactor": 0.0
            }
        }
    ]
```

AdamMitchell-ms:
Fixed in latest.

PR #252, opened 2018-09-27
Fix for roughness factor issue #251

tolen23:
Proposed fix for issue #251.

**Explanation:** 
In the base class [MetalRough2StandardMap](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRough2StandardMap.cs) RoughnessFactor is statically set to 0.5 since the standard shader does not support this parameter:
```
// not supported by the Standard shader
public virtual double RoughnessFactor
{
	get { return 0.5; }
	set { return; 
}
```
 The inherited class [MetalRoughMap](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs) does not override this, so we have the same behavior here, i.e. the original RoughnessFactor that was set in the gltf file gets lost.

However, in class [MetalRoughMap](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs) the default shader is PbrMetallicRoughness. When looking at line 12 in [PbrMetallicRoughness.shader](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/Assets/UnityGLTF/Shaders/PbrMetallicRoughness.shader)  we can see the following:

`_Glossiness("Roughness Factor", Range(0.0, 1.0)) = 1.0`

This means that
1. Glossiness/Roughness Factor is supported by this shader
2. The default value is 1.0

I guess this is the reason why we always get 1.0 in the end, not 0.5 (see description of issue #251).

In my opinion this issue can be resolved by overriding RoughnessFactor in class MetalRoughMap as proposed in this pull request.

However, I'm not sure if we would have to deal with the case that a shader is used which does not support the property (PbrMetallicRoughness is the default shader, but other shaders can be specified through other constructors).

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs
@@ -57,5 +57,15 @@ namespace UnityGLTF
 			base.Copy(copy);
 			return copy;
 		}
+		
+		public override double RoughnessFactor
+		{
+			get { 
+				return _material.GetFloat("_Glossiness");
+			}
+			set { 
+				_material.SetFloat("_Glossiness", (float)value);
+			}
+		}
 	}
 }


blgrossMS:
Looks good to me. @stevenvergenz  or @JohnCopicMS do you have any opinions?

JohnCopicMS:
Seems good to me as well. I assume when I override the shaders that I am responsible for implementing any properties like this that I care about.

UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs
```diff
@@ -57,5 +57,15 @@ public override IUniformMap Clone()
 			base.Copy(copy);
 			return copy;
 		}
+		
+		public override double RoughnessFactor
```
stevenvergenz: Shouldn't this be of type `float`?

tolen23: My proposal is inspired by the definition of MetallicFactor in class MetalRough2StandardMap, where you can find the same mix of ```double``` and ```float```:

```
public virtual double MetallicFactor
{
	get { return _material.GetFloat("_Metallic"); }
	set { _material.SetFloat("_Metallic", (float)value); }
}
```
However, you are right, on a second look this seems a bit odd, but I don't know if it has any impact in practice (or maybe is even done in this way by intention).

The same can also be found in other classes, e.g. SpecGloss2StandardMap:

```
public virtual double GlossinessFactor
{
	get { return _material.GetFloat("_GlossMapScale"); }
	set
	{
		_material.SetFloat("_GlossMapScale", (float)value);
		_material.SetFloat("_Glossiness", (float)value);
	}
}
```

stevenvergenz: Given that the materials only support single-precision uniforms, these should all probably be floats and not doubles.

AdamMitchell-ms:
Looking at old pulls.  It looks like this same change got made by someone else.

PR #259, opened 2018-10-01
Proposal for new architecture for UnityGLTF library

blgrossMS:


vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,181 @@
...
+        );
+
+    /// <summary>
+    /// Creates a Unity Texture2D from a glTF texture
+    /// </summary>
+    /// <param name="unityGLTFFObject">Object which contains information to parse</param>
+    /// <param name="textureId">Texture to load from glTF object.</param>        
+    /// <returns>The created Unity object</returns>
+    public Task<Texture2D> ImportTextureAsync(
+        UnityGLTFObject unityGLTFFObject,
```
StephenHodgson: `unityGLTFFObject` has two FFs

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,181 @@
...
+        );
+
+    /// <summary>
+    /// Creates a Unity GameObject from a glTF node
+    /// </summary>
+    /// <param name="unityGLTFFObject">Object which contains information to parse</param>
+    /// <param name="nodeId">Node of the glTF object to load.</param>        
+    /// <returns>The created Unity object</returns>
+    public Task<GameObject> ImportNodeAsync(
+        UnityGLTFObject unityGLTFFObject,
```
StephenHodgson: `unityGLTFFObject` has two FFs

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,181 @@
...
+    ContinueDefaultExecution,
+    Handled
+}
+
+// Example calling pattern:
+public async void LoadGLBs()
+{
+    UnityGLTFObject sampleObject = new UnityGLTFObject("http://samplemodels/samplemodel.glb");
+    UnityGLTFObject boxObject = new UnityGLTFObject("http://samplemodels/box.glb");
+    IDataLoader dataLoader = new WebRequestLoader();
```
StephenHodgson: Couldn't we just read the binary data asynchronously from disk?

blgrossMS: Yes you can specify a FileLoader if desired

vNext/UnityGLTF_vnext_proposal.md
```diff
@@ -0,0 +1,39 @@
...
+
+Due to these issues I am proposing a breaking API change, but the goal is for it to be a stable repository after that.
+
+My proposal for the new system would also remove support for any Unity version that does not properly support async/await (which means supporting the new language features + Unity having a sync context to ensure return to main thread).
+
+## Simpler API
+
+The first change I would like to make is to simplify the API. Currently the number of constructors and configuration options is overwhelming in creating a GLTFSceneImporter.
+
+    public static class UnityGLTFLoader
```
ryantrem: This block of code is out of date... delete?

vNext/UnityGLTFv2.cs
stevenvergenz: Let's not assume disk access. We're pulling it from the network.

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,181 @@
+/// <summary>
+/// Handles importing object from schema into Unity and handles exporting of objects from Unity into schema
+/// </summary>
+public partial class UnityGLTFImporter
+{
+    public UnityGLTFImporter(
+        IDataLoader dataLoader,
```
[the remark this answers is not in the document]
blgrossMS: I like it being an interface because then we can easily provide default implementation. I suppose we could do the same with default classes but this seems cleaner.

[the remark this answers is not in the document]
blgrossMS: I also envision that someone will make a class like 

`public class WebRequestResolver : IDataLoader, IDataWriter`

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,181 @@
+/// <summary>
+/// Handles importing object from schema into Unity and handles exporting of objects from Unity into schema
+/// </summary>
+public partial class UnityGLTFImporter
```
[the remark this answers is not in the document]
StephenHodgson: Agreed. Partial classes are also kind of hard to hunt down as well.

vNext/UnityGLTFv2.cs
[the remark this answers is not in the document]
ryantrem: Default implementations of a Func could easily be provided as static methods of a static class. However, if you expect it to be common to share logic across the IDataLoader and IDataWriter, or if you expect the implementation to be stateful, then I feel convinced that an interface is reasonable.

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,181 @@
...
+        IDataLoader dataLoader,
+        UnityGLTFImporterConfig config = new UnityGLTFImporterConfig()
+        );
+
+    /// <summary>
+    /// Creates a Unity GameObject from a glTF scene
+    /// </summary>
+    /// <param name="unityGLTFFObject">Object which contains information to parse</param>
+    /// <param name="sceneId">Scene of the glTF to load</param>
+    /// <returns>The created Unity object</returns>
```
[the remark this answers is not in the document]
blgrossMS: Added

vNext/UnityGLTFv2.cs
```diff
@@ -1,63 +1,70 @@
...
-    private Task<Texture2D> ConstructTexture();
+    private virtual Task<Texture2D> ConstructTexture();
 }
 
-public class UnityGLTFImporterConfig
+public class ImporterConfig
 {
-    public UnityGLTFImporterConfig();
-    public UnityGLTFImporterConfig(GLTFExtensionRegistry registry, GLTFImportOptions importOptions);
+    public ImporterConfig();
```
ryantrem: Why do we need this paramterless constructor?

vNext/UnityGLTFv2.cs
```diff
@@ -66,7 +73,7 @@ public class UnityGLTFImporterConfig
 /// </summary>
 public class IDataLoader
```
ryantrem: interface, not class
Also wondering about the name of this interface... I wonder if something like IResourceLocator would be more self explanatory? Not sure...

vNext/UnityGLTFv2.cs
```diff
@@ -66,7 +73,7 @@ public class UnityGLTFImporterConfig
 /// </summary>
 public class IDataLoader
 {
-    Task<Stream> LoadStream(string uri);
+    Task<Stream> LoadStreamAsync(string uri, CancellationToken ct = CancellationToken.None);
```
ryantrem: How about `Uri uri` instead of `string uri`. Just require at the contract level that a valid Uri is passed in.

vNext/UnityGLTFv2.cs
blgrossMS: The default constructor is used for when there are no options that are being set

vNext/UnityGLTFv2.cs
blgrossMS: Because it's not necessarily a web uri, which seems to be what the uri class is built around. Maybe path is a better parameter name?

vNext/UnityGLTFv2.cs
```diff
@@ -149,9 +159,34 @@ public interface IUnityGLTFExtension
...
-    Task<ExtensionReturnObject<GameObject>> CreateNodeAsync(Node gltfNode, int indexToLoad);
-    Task<ExtensionReturnObject<MeshPrimitive>> CreateMeshPrimitiveAsync(GLTFMesh mesh, ILoaderContext loaderContext, int indexToLoad);
+    /// <summary>
+    /// Creates a Unity object out of the glTF schema object
+    /// </summary>
+    /// <param name="importer">The importer that is used to load the object</param>
+    /// <param name="unityGLTFObject">Object that is being loaded</param>
+    /// <param name="sceneId">Index object which resolves to object in GLTFRoot</param>
+    /// <returns>The loaded glTF scene as a GameObject hierarchy</returns>
+    Task<ExtensionReturnObject<GameObject>> CreateSceneAsync(Importer importer, UnityGLTFObject unityGLTFObject, GLTF.Schema.SceneId sceneId);
```
ryantrem: Why do we need the Importer passed in here? Is it to allow the method to call the default load logic for a scene or something? If so, we should probably pass in an interface with the intended subset of the functionality. Otherwise, it is confusing how you are supposed to use this from the context of an extension (unless it is really expected that you might call *any* method on the importer from this context). Alternatively, we could restrict this further to only having the ability to invoke the base functionality for the type in question (e.g. from this context, you could call just one method for default scene loading, and in CreateNodeAsync you could call just one method for default node loading, etc.).

ryantrem: Why is the UnityGLTFObject passed in here? Is it so the extension can basically cache data in it? If so, then I guess this implies that the UnityGLTFObject is not opaque from the standpoint of the consumer of this API?

vNext/UnityGLTFv2.cs
```diff
@@ -175,7 +209,8 @@ public async void LoadGLBs()
     UnityGLTFObject boxObject = new UnityGLTFObject("http://samplemodels/box.glb");
```
ryantrem: Is this assuming that deserialization is synchronous? Seems like this would naturally lead people down a bad path.

vNext/UnityGLTFv2.cs
```diff
@@ -1,63 +1,70 @@
...
-    /// <param name="unityGLTFFObject">Object which contains information to parse</param>
+    /// <param name="unityGLTFObject">Object which contains information to parse</param>
     /// <param name="sceneId">Scene of the glTF to load</param>
+    /// <param name="progress">Progress of load completion</param>
     /// <returns>The created Unity object</returns>
-    public Task<GameObject> ImportSceneAsync(
-        UnityGLTFObject unityGLTFFObject,
-        int sceneId = -1
+    public virtual Task<GameObject> ImportSceneAsync(
+        UnityGLTFObject unityGLTFObject,
```
ryantrem: The whole UnityGLTFObject stateful wrapper still seems like a confusing part of this design to me. I think I'm now leaning more towards just passing in the raw deserialized data structure for the gltf root, and additionally (optionally) passing in an AssetCache by ref. Any problems with that approach?

vNext/UnityGLTFv2.cs
stevenvergenz: The [URI documentation](https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netframework-4.7.2) says that it supports the `file://` protocol, which in turn supports relative URIs. Seems like that would be enough.

vNext/UnityGLTFv2.cs
stevenvergenz: Perhaps this layer of intention could be implemented as an IExtensionContext interface instance passed into the function, where Importer implements IExtensionContext?

blgrossMS: Yes that UnityGLTFObject has the cache of data inside which may be needed for the extension

vNext/UnityGLTFv2.cs
blgrossMS: No, this is just passing in a uri that will be resolved during the actual load

vNext/UnityGLTFv2.cs
blgrossMS: It doesn't enforce that asset caches and paired to a glTF object. It's just decoupling the two rather than having them stored in an object. I should update the design to have UnityGLTFObject expose the AssetCache.

vNext/UnityGLTFv2.cs
blgrossMS: It's not really locating resources. It's only resolving them to a stream.

vNext/UnityGLTFv2.cs
blgrossMS: This mirrors the functionality of Babylon. An extension may want to call any of the functions on the loader. I suppose there is a limit to what extension could want to do (an image load is not going to probably edit the animation hierarchy). I'm not sure how an interface would look different than the internal loader functions.

One thought is that the `Importer` interface is publicly created and there is an `ImporterImpl` which is not able to be publicly constructed but all methods are publicly accessible

blgrossMS: That's right. The extension may need additional data from either the root or the asset cache.

ryantrem: Pass the CancellationToken to the gltf extension methods as well.

vNext/UnityGLTFv2.cs
ryantrem: Why not just pass in null if you don't want to set options? It seems weird that the options are to either provide extensions *and* options, *or* nothing.

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,238 @@
...
+public class ExporterConfig
+{
+    public ExporterConfig();
+    public ExporterConfig(List<IUnityExtensionFactory> registry, GLTFExportOptions exportOptions);
+}
+
+public class GLTFExportOptions
+{
+    /// <summary>Whether to write the object out as a GLB</summary>
+    bool ShouldWriteGLB { get; set ; }
```
ryantrem: Does it not make more sense for this to just be a parameter of the actual Export* method(s)? Is it not possible to have a single exporter instance that exports both gltf and glb?

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/// <summary>
+/// Unity glTF extension wrapper
+/// Base class does returns continuation behavior of not handled
+/// </summary>
+public abstract class UnityExtensionFactoryBase
```
ryantrem: Why is the word "Factory" part of this name?

vNext/UnityGLTFv2.cs
```diff
@@ -0,0 +1,238 @@
...
+    T ReturnObject;
+}
+
+public enum ExtensionContinuationBehavior
+{
+    NotHandled,
+    Handled
+}
+
+// Example calling pattern:
```
ryantrem: No longer valid, correct?

blgrossMS:
Active work has slowed down on this. We are more taking the approach now of slowly changing the existing code to more closely match the proposed architecture. Anyone who wants to help out can!

hybridherbst:
Closed – outdated. UnityGLTF architecture has been adjusted over the years, and now features a flexible import/export plugin architecture.

PR #266, opened 2018-10-05
Fixes from previous PR around glb and bug fixes

blgrossMS:
- Verified that JSON is deserialized correctly
- Fixed spacing


dave-hill:
It would be good to mention which bugs were fixed in the commit message, or link the associated issues.

UnityGLTF/Assets/UnityGLTF/Shaders/UnityStandardInput.cginc
```diff
@@ -13,78 +13,78 @@
...
 
 #if (_DETAIL_MULX2 || _DETAIL_MUL || _DETAIL_ADD || _DETAIL_LERP)
-    #define _DETAIL 1
+	#define _DETAIL 1
 #endif
 
 //---------------------------------------
-half4       _Color;
-half        _Cutoff;
+half4	   _Color;
```
dave-hill: I think a tool has done something strange here - this is an even weirder mix of tabs and spaces!

PR #288, opened 2018-11-09
Fixed loading performance by removing most yields and adding it a performance based yield

blgrossMS:
- Fixed loading performance by removing most yields and adding it a performance based yield


UnityGLTF/Assets/UnityGLTF/Scripts/Async/AsyncCoroutineHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Threading.Tasks;
+using UnityEngine;
+namespace UnityGLTF
+{
+    public interface IAsyncCoroutineHelper
+    {
+        Task RunAsTask(IEnumerator coroutine, string name);
+    }
+
+	public class AsyncCoroutineHelper : MonoBehaviour, IAsyncCoroutineHelper
```
ryantrem: There is a bunch of weird indentation/formatting in this file.

UnityGLTF/Assets/UnityGLTF/Scripts/Async/AsyncCoroutineHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Threading.Tasks;
+using UnityEngine;
+namespace UnityGLTF
+{
+    public interface IAsyncCoroutineHelper
+    {
+	   Task RunAsTask(IEnumerator coroutine, string name);
+    }
+
+	public class AsyncCoroutineHelper : MonoBehaviour, IAsyncCoroutineHelper
```
ryantrem: What are the responsibilities of this class? It looks like it both lets you "start" a coroutine from any thread, and also "converts" it into a Task. Is that right?

PR #325, opened 2018-12-13
Add error handling to scene loading, refactor WebRequestLoader

stevenvergenz:
This PR adds an exception argument to the LoadScene callbacks, which are invoked even if the load/parse/import process fails.

It also refactors WebRequestLoader to use System.Net.Http instead of UnityWebRequest. This way we don't have to convert to and from Coroutines, but run directly with Tasks. It has the added benefit of playing better with the above exception handling system.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -159,7 +159,7 @@ namespace UnityGLTF
 		/// <param name="sceneIndex">The scene to load, If the index isn't specified, we use the default index in the file. Failing that we load index 0.</param>
 		/// <param name="onLoadComplete">Callback function for when load is completed</param>
 		/// <returns></returns>
-		public async Task LoadSceneAsync(int sceneIndex = -1, Action<GameObject> onLoadComplete = null)
+		public async Task LoadSceneAsync(int sceneIndex = -1, Action<GameObject, Exception> onLoadComplete = null)
 		{
 			try
 			{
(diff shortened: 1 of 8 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -159,7 +159,7 @@ public GameObject LastLoadedScene
 		/// <param name="sceneIndex">The scene to load, If the index isn't specified, we use the default index in the file. Failing that we load index 0.</param>
 		/// <param name="onLoadComplete">Callback function for when load is completed</param>
 		/// <returns></returns>
-		public async Task LoadSceneAsync(int sceneIndex = -1, Action<GameObject> onLoadComplete = null)
+		public async Task LoadSceneAsync(int sceneIndex = -1, Action<GameObject, Exception> onLoadComplete = null)
```
ryantrem: I think it would make more sense for this to be an ExceptionDispatchInfo.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -178,10 +178,16 @@ public async Task LoadSceneAsync(int sceneIndex = -1, Action<GameObject> onLoadC
...
 					await LoadJson(_gltfFileName);
 				}
+
 				await _LoadScene(sceneIndex);
 
 				Cleanup();
 			}
+			catch (Exception ex)
+			{
+				onLoadComplete?.Invoke(null, ex);
```
ryantrem: You can get an ExceptionDispatchInfo via ExceptionDispatchInfo.Capture(ex).

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
stevenvergenz: Didn't know about this class. Updated!

PR #333, opened 2019-01-03
Pass "isGLB = true" when writing GLB

iBicha:
GLB's json should not have indentation


blgrossMS:
Thanks for the contribution!

PR #342, opened 2019-01-09
Fixed bug where indices were not always getting their buffer loaded

blgrossMS:


martinken:
It would be great to get this merged as I ran into the same issue and this MR fixed it. +1

PR #354, opened 2019-01-27
Fix play mode tests

blgrossMS:
- Fixed tests
- Moved their location
- Removed old unity test tools framework 
- GLTFComponent now plays animations


blgrossMS:
@azakariaMSFT take a look

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFComponent.cs
```diff
@@ -122,6 +132,15 @@ public async Task Load()
 						renderer.sharedMaterial.shader = shaderOverride;
 					}
 				}
+
+				if (PlayAnimationOnLoad)
+				{
+					Animation[] animations = sceneImporter.LastLoadedScene.GetComponents<Animation>();
+					foreach (Animation anim in animations)
```
azakariaMSFT: I suggest having a property for choosing a specific animation index. Play all animations can be overwhelming

blgrossMS: There is no linkage here between animations and their indices. This is something we may want to add in the future

PR #359, opened 2019-02-14
Removed redundant ref count increment on duplication.

CharlesWanMS:
Remove redundant ref count increment on cachedData on duplication.
The ref count on CachedData is already incremented once within the setter of newGltfObjectComponent.CachedData.


jiamingfeng:
How do I pass the AppVeyor build?

> Build started
git clone -q https://github.com/KhronosGroup/UnityGLTF.git C:\projects\unitygltf
Filtering content: 100% (8/8), 13.40 MiB | 12.25 MiB/s, done.
git fetch -q origin +refs/pull/359/merge:
git checkout -qf FETCH_HEAD
**Specify which project or solution file to use because the folder contains more than one project or solution file**.

PR #364, opened 2019-02-19
add the ability to control the final LOD when far away (cull vs render last lod)

azakariaMSFT:
add the ability to control whether objects with lod should render when very far away


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -96,7 +96,12 @@ public enum ColliderType
 
 		public bool KeepCPUCopyOfMesh = true;
 
-		protected struct GLBStream
+		/// <summary>
+		/// When screen coverage is above threashold and no LOD mesh cull the object
+		/// </summary>
+		public bool CullFarLOD = false;
+
+        protected struct GLBStream
```
blgrossMS: spaces

Issue #357, opened 2019-01-31
WebGL run fails with failed thread creation


FreakTheMighty:
I've built from UnityGLTF from master, added the `GLTFComponent`to a scene to load a remote GLTF and disabled multithreading. When I built to WebGL I get the following error. I've found that the 1.0.0 release builds and runs in webgl.

Related: #138 

```
SystemException: Thread creation failed.
UnityLoader.js:1043   at System.Threading.Thread.StartInternal (System.Security.Principal.IPrincipal principal, System.Threading.StackCrawlMark& stackMark) [0x00000] in <00000000000000000000000000000000>:0 
UnityLoader.js:1043 Rethrow as TypeInitializationException: The type initializer for 'System.Threading.Timer.Scheduler' threw an exception.
UnityLoader.js:1043 
UnityLoader.js:1043 Rethrow as TypeInitializationException: The type initializer for 'System.Threading.Timer' threw an exception.
UnityLoader.js:1043   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x00000] in <00000000000000000000000000000000>:0 
UnityLoader.js:1043 --- End of stack trace from previous location where exception was thrown ---
UnityLoader.js:1043   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x00000] in <00000000000000000000000000000000>:0 
UnityLoader.js:1043 --- End of stack trace from previous location where exception was thrown ---
UnityLoader.js:1043   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x00000] in <00000000000000000000000000000000>:0 
UnityLoader.js:1043 --- End of stack trace from previous location where exception was thrown ---
UnityLoader.js:1043   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x00000] in <00000000000000000000000000000000>:0 
UnityLoader.js:1043 --- End of stack trace from previous location where exception was thrown ---
UnityLoader.js:1043   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x00000] in <00000000000000000000000000000000>:0 
```

FreakTheMighty:
Apologies, forgot platform information:

Unity 2018.3.1f1
OS X: 10.14.2
master @ bb189f382697317321a583b22f0aab3692414366

kolodi:
I've just tried and I have the same thread creation errors.
Unity 2018.3.3f1
[...]

PR #367, opened 2019-02-26
WebGL download fix

kolodi:
On WebGL we can't use multithreading as mentioned in #357. Apart from diabling the "Multithreaded" option for the GLTF Component there was a problem with .NET http client which is using mrlthithreading as well. I nthis fix I have introducing of using Unity's standard `UnityWebRequest`, for now, only for WebGL build target.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,6 @@ _pkginfo.txt
 !*.[Cc]ache/
 
 *.orig
+
+# Unity specific
+ProjectSettings/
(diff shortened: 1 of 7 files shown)


stevenvergenz:
Thanks for the PR! We will be reviewing this as soon as we can.

yaakovyitzchak:
@stevenvergenz hey im not able to build with GLTF and webgl at all https://github.com/KhronosGroup/UnityGLTF/issues/436 153 errors?

PR #371, opened 2019-03-07
Update KHR_texture_transform support

stevenvergenz:
* Update `EXT_texture_transform` extension name to `KHR_`
* Add texture transform `rotation` field
* Add texture transform properties to IUniformMap and implementations


blgrossMS:
@stevenvergenz this looks good, but doesn't have any shader changes associated with rotation

PR #383, opened 2019-03-27
Invert the _Glossiness Material value

DavidArayan:
The _Glossiness value is mapped into glTF roughnessFactor value. This causes glossy materials in Unity3D to appear matte in ThreeJS viewer and vice-versa. Inverting this value fixes the issue.


blgrossMS:
Thank you for the contribution!

PR #407, opened 2019-04-16
Add Roughness float values to MetalRoughMap.cs

rferrese:
If the GLTF has a float value for the roughness, set it in the texture

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs
@@ -86,6 +86,12 @@ namespace UnityGLTF
 			set { return; }
 		}
 
+		public override double RoughnessFactor
+		{
+			get { return _material.GetFloat("_Glossiness"); }
+			set { _material.SetFloat("_Glossiness", (float)value); }
+		}
+
 		public override IUniformMap Clone()
 		{
 			var copy = new MetalRoughMap(new Material(_material));


UnityGLTF/Assets/UnityGLTF/Scripts/UniformMaps/MetalRoughMap.cs
```diff
@@ -86,6 +86,12 @@ public override int MetallicRoughnessXTexCoord
 			set { return; }
 		}
 
+		public override double RoughnessFactor
+		{
+			get { return _material.GetFloat("_Glossiness"); }
+			set { _material.SetFloat("_Glossiness", (float)value); }
+		}
```
AdamMitchell-ms: Does this render correctly?  Glossiness is the inverse of roughness.  I do see that MetallicRoughnessTexture in the cs files refers to MetallicGlossMap in the shaders.  Is this an optimization to use the same variable to hold different values for the different shader variants?  It seems odd to me.

rferrese: It is very odd.  Im not sure exactly what the deal is.  In the PbrMetallicRoughness.shader file, it has
`_Glossiness("Roughness Factor", Range(0.0, 1.0)) = 1.0`
as the shader parameter.  I tested in babylonjs (see pictures) and it renders the same in the editor with the change.

Best I can tell, is that PbrMetallicRoughness.shader and PbrSpecularGlossiness.shader both use UnityStandardInput.cginc, which uses _Glossiness.  These shaders also use a couple unity internal cginc files, which I assume require the _Glossiness parameter as well.  Though without digging through it, I can only speculate.

But, like I said, it works.

babylonjs (left) vs Unity (right)
image

PR #408, opened 2019-04-16
Add names to textures created by GLTFSceneImporter

rferrese:
Add names to the textures to track memory leaks

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -630,7 +630,7 @@ namespace UnityGLTF
 		protected virtual async Task ConstructUnityTexture(Stream stream, bool markGpuOnly, bool isLinear, GLTFImage image, int imageCacheIndex)
 		{
 			Texture2D texture = new Texture2D(0, 0, TextureFormat.RGBA32, true, isLinear);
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -630,6 +630,7 @@ protected async Task ConstructImage(GLTFImage image, int imageCacheIndex, bool m
 		protected virtual async Task ConstructUnityTexture(Stream stream, bool markGpuOnly, bool isLinear, GLTFImage image, int imageCacheIndex)
 		{
 			Texture2D texture = new Texture2D(0, 0, TextureFormat.RGBA32, true, isLinear);
+			texture.name = nameof(GLTFSceneImporter);
```
AdamMitchell-ms: Instead of naming it "GLTFSceneImporter", set texture.name to image.name.  Not sure that there will be an actual name in there very often, but I'm also not sure how useful it is to create a bunch of textures with the same name.

Or maybe some combination?
texture.name = nameof(GLTFSceneImporter) + "." + image.name;

rferrese: Really this was just for attributing texture allocations to the GLTF addon.  I can't think of any time when I've used the texture name for anything.  But it certainly doesn't hurt to have the image name as well

PR #409, opened 2019-04-16
Fix texture memory leak

rferrese:
- Change MaterialCacheData, MeshCacheData, and TextureCacheData to IDisposable
- Add ImageCache array to RefCountedCacheData
- Dont allow replacing arrays in RefCountedCacheData after construction to prevent leaks
- Prevent recreation of TextureCacheData (leaking references to cloned textures)
- Add debug asserts to texture reference setting to ensure no leaked references

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/Cache/RefCountedCacheData.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/Cache/RefCountedCacheData.cs
@@ -112,7 +112,7 @@ namespace UnityGLTF.Cache
 				MaterialCache[i] = null;
 			}
 
-			// Destroy the cached textures
+			// Destroy the cached images
 			for (int i = 0; i < ImageCache.Length; i++)
 			{
 				if (ImageCache[i] != null)
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/Cache/RefCountedCacheData.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/Cache/RefCountedCacheData.cs
@@ -25,17 +25,30 @@ namespace UnityGLTF.Cache
 		/// <summary>
 		/// Meshes used by this GLTF node.
 		/// </summary>
-		public MeshCacheData[][] MeshCache { get; set; }
+		public MeshCacheData[][] MeshCache { get; private set; }
 
 		/// <summary>
 		/// Materials used by this GLTF node.
 		/// </summary>
-		public MaterialCacheData[] MaterialCache { get; set; }
+		public MaterialCacheData[] MaterialCache { get; private set; }
 
 		/// <summary>
 		/// Textures used by this GLTF node.
 		/// </summary>
-		public TextureCacheData[] TextureCache { get; set; }
+		public TextureCacheData[] TextureCache { get; private set; }
+
+		/// <summary>
+		/// Textures from the AssetCache that might need to be cleaned up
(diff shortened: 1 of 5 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/Cache/RefCountedCacheData.cs
```diff
@@ -80,28 +93,32 @@ private void DestroyCachedData()
...
 
 			// Destroy the cached materials
 			for (int i = 0; i < MaterialCache.Length; i++)
 			{
-				if (MaterialCache[i] != null)
+				MaterialCache[i]?.Dispose();
+				MaterialCache[i] = null;
+			}
+
+			// Destroy the cached textures
```
AdamMitchell-ms: nit: images, not textures

```suggestion
			// Destroy the cached images
```

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -439,10 +440,13 @@ protected async Task ConstructImageBuffer(GLTFTexture texture, int textureIndex)
...
+			if (_assetCache.TextureCache[textureIndex] == null)
 			{
-				TextureDefinition = texture
-			};
+				_assetCache.TextureCache[textureIndex] = new TextureCacheData
+				{
+					TextureDefinition = texture
+				};
+			}
 		}
```
AdamMitchell-ms: Should this throw in the else case?  Or warn or something?  Or is it ok to silently do nothing on multiple calls to this function with the same textureIndex?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -453,19 +457,19 @@ protected IEnumerator WaitUntilEnum(WaitUntil waitUntil)
 		private async Task LoadJson(string jsonFilePath)
 		{
 #if !WINDOWS_UWP
-			 if (IsMultithreaded && _loader.HasSyncLoadMethod)
```
AdamMitchell-ms: Thanks for all the little whitespace fixes

UnityGLTF/Assets/UnityGLTF/Scripts/Cache/RefCountedCacheData.cs
AdamMitchell-ms: nit: images, not textures

```suggestion
			// Destroy the cached images
```

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
rferrese: Really this entire method needs work.  From what it looks like, this will fill in either ImageStreamCache or BufferCache depending on what type it is, but the IF statements dont make that intention clear.  Both of these are guarded against multiple calls with the same ID.  This also new's TextureCacheData, but that wasnt guarded (which was causing duplicate textures to be created).  I'm not entirely sure why this function gets called multiple times, but subsequent calls with the same ID do in fact intend to populate with the same values.

As for your comment, this frequently gets called multiple times with the same ID, so a warning would probably not be the right thing to do in this case.  If we rewrote this so that it only got called once per texture, I would be all for it.  (similar to the debug asserts I added elsewhere)

Issue #373, opened 2019-03-12
Cancellation not supported.


sanjeev3094:
Loading a scene cannot be canceled. 
**Usecase**: Suppose I load a scene and then want to cancel the load operation(due to some requirement) then it cannot be done as cancellation tokens are not accepted as a parameter.
It would be great if there can be a support for cancellation.

tufeixp:
I have added a token for exit, but I'm not sure if I can upload it.

MephestoKhaan:
Not sure if that token is working for me? Used LoadAsync as a coroutine and set the token, and even tho the load was interrupted the memory did not go down.

pfcDorn:
Please update to the latest version and open a new issue if the problem persists. Thanks!

PR #412, opened 2019-04-17
Add task cancellation support

rferrese:
Fixes issue: #373 (Cancellation not supported)

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -1283,6 +1283,8 @@ namespace UnityGLTF
 
 		protected virtual async Task ConstructMesh(GLTFMesh mesh, Transform parent, int meshId, Skin skin, CancellationToken cancellationToken)
 		{
+			cancellationToken.ThrowIfCancellationRequested();
+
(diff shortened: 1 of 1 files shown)


rferrese:
For reference, here is the call hierarchy.

ThrowIfCancellationRequested has been added to:
- LoadSceneAsync
- _LoadNode
- ConstructBufferData
- ConstructNode
- ConstructMesh

image

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1281,6 +1290,8 @@ protected virtual async Task ConstructMesh(GLTFMesh mesh, Transform parent, int
 
 			for (int i = 0; i < mesh.Primitives.Count; ++i)
 			{
+				cancellationToken.ThrowIfCancellationRequested();
+
 				var primitive = mesh.Primitives[i];
 				int materialIndex = primitive.Material != null ? primitive.Material.Id : -1;
```
AdamMitchell-ms: For the two awaited calls below (ConstructMeshPrimitive and SetupBones), you should either call ThrowIfCancellationRequested after calling them, or pass the cancellationToken in to them and have them ThrowIfCancellationRequested after each of their awaits.

PR #415, opened 2019-04-24
Let Unity move the Integration meta file to the correct location.

AdamMitchell-ms:
Also clean up some loader settings in TestScene.unity.

Addresses issue #405


AdamMitchell-ms:
I just realized that issue #405 is already addressed by pull #406.  So I'll look at getting that PR in instead.

PR #422, opened 2019-05-03
OrbitCamera now supports right mouse buttons to change position

AdamMitchell-ms:


UnityGLTF/Assets/UnityGLTF/Examples/OrbitCameraController.cs
```diff
@@ -44,38 +52,52 @@ void Start()
...
+			transform.rotation = rotation;
+			transform.position = position;
+
+			prevMousePosition = Input.mousePosition;
+		}
+
+		private Vector3 ProjectScreenPointToTargetPlane(Vector3 screenPosition)
+		{
+			var ray = camera.ScreenPointToRay(screenPosition);
+			var planeDistance = distance / Mathf.Cos(Mathf.Deg2Rad * Vector3.Angle(camera.transform.forward, ray.direction));
```
pjramirez-ms: Do you need to handle divide by zero here?

AdamMitchell-ms: It's not possible for this to be zero with a perspective camera because the angles between the ray and the camera's forward can never quite reach 90 degrees.

PR #424, opened 2019-05-14
Add ImportProgress indication

rferrese:
First pass of adding some indication of the nodes/textures that are being loaded

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -46,15 +46,35 @@ namespace UnityGLTF
 		public BoneWeight[] BoneWeights;
 	}
 
-	public class ImportProgress
+	public struct ImportProgress
 	{
-		public bool IsDownloaded = false;
+		public bool IsDownloaded;
 
-		public int NodeTotal = 0;
-		public int NodeLoaded = 0;
+		public int NodeTotal;
+		public int NodeLoaded;
 
-		public int TextureTotal = 0;
-		public int TextureLoaded = 0;
+		public int TextureTotal;
+		public int TextureLoaded;
+
+		public int BuffersTotal;
+		public int BuffersLoaded;
+
+		public float Progress
+		{
+			get
+			{
+				int total = NodeTotal + TextureTotal + BuffersTotal;
+				int loaded = NodeLoaded + TextureLoaded + BuffersLoaded;
+				if (total > 0)
+				{
+					return (float)loaded / total;
+				}
+				else
+				{
+					return 0.0f;
+				}
+			}
+		}
 	}
 
 	/// <summary>
@@ -160,8 +180,9 @@ namespace UnityGLTF
 		protected ILoader _loader;
 		protected bool _isRunning = false;
 
-		public ImportProgress Progress { get; } = new ImportProgress();
-		
+		protected ImportProgress progressStatus = default(ImportProgress);
+		protected IProgress<ImportProgress> progress = null;
+
 		/// <summary>
 		/// Creates a GLTFSceneBuilder object which will be able to construct a scene based off a url
 		/// </summary>
@@ -210,7 +231,7 @@ namespace UnityGLTF
 		/// <param name="onLoadComplete">Callback function for when load is completed</param>
 		/// <param name="cancellationToken">Cancellation token for loading</param>
 		/// <returns></returns>
-		public async Task LoadSceneAsync(int sceneIndex = -1, bool showSceneObj = true, Action<GameObject, ExceptionDispatchInfo> onLoadComplete = null, CancellationToken cancellationToken = default(CancellationToken))
+		public async Task LoadSceneAsync(int sceneIndex = -1, bool showSceneObj = true, Action<GameObject, ExceptionDispatchInfo> onLoadComplete = null, CancellationToken cancellationToken = default(CancellationToken), IProgress<ImportProgress> progress = null)
 		{
 			try
 			{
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -149,7 +160,8 @@ protected struct GLBStream
 		protected ILoader _loader;
 		protected bool _isRunning = false;
 
-
+		public ImportProgress Progress { get; } = new ImportProgress();
```
AdamMitchell-ms: Instead of a public Progress, you should:

Take in an optional IProgress<ImportProgress> progress in LoadSceneAsync() and use that to report progress.

Then the ImportProgress should probably also change to make all of its public fields become public getter-only properties that are set in the constructor.  So you construct one then call IProgress.Report();

Or make ImportProgress a struct so that it is passed by value.  Basically I don't want it to be possible for the caller to accidentally (or maliciously) set the progress value on us.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -46,6 +46,17 @@ public class UnityMeshData
...
+	public class ImportProgress
+	{
+		public bool IsDownloaded = false;
+
+		public int NodeTotal = 0;
+		public int NodeLoaded = 0;
+
+		public int TextureTotal = 0;
+		public int TextureLoaded = 0;
+	}
```
AdamMitchell-ms: Should you also include mesh data?  (aka. buffers)

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -664,6 +710,7 @@ protected virtual async Task ConstructUnityTexture(Stream stream, bool markGpuOn
 				if (_asyncCoroutineHelper != null) await _asyncCoroutineHelper.YieldOnTimeout();
 				//	NOTE: the second parameter of LoadImage() marks non-readable, but we can't mark it until after we call Apply()
 				texture.LoadImage(buffer, markGpuOnly);
+				Progress.TextureLoaded++;
```
AdamMitchell-ms: Instead of two increments in the if/else block in this function, you could have a single increment at the end of the if/else blocks.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
AdamMitchell-ms: Consider adding a public float Progress with a getter that looks at the loaded vs totals and returns a progress percentage.  It would be a helper function for callers.  I assume you'd need to write that logic to drive a progress bar somewhere anyway.  Maybe it could be a nullable float where it's null (presented to user as indeterminate progress) until isDownloaded is true?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -573,6 +588,36 @@ protected async Task _LoadScene(int sceneIndex = -1, bool showSceneObj = true, C
 			_lastLoadedScene = CreatedObject;
 		}
 
+		private void GetGtlfContentTotals(GLTFScene scene)
+		{
+			// Count Nodes
+			Queue<Node> nodeQueue = new Queue<Node>();
+
+			// Add scene nodes
+			for (int i = 0; i < scene.Nodes.Count; ++i)
```
pjramirez-ms: Probably want to check for null "Nodes" object

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1943,6 +1991,7 @@ protected virtual void ConstructImageFromGLB(GLTFImage image, int imageCacheInde
 			bufferContents.Stream.Position = bufferView.ByteOffset + bufferContents.ChunkOffset;
 			bufferContents.Stream.Read(data, 0, data.Length);
 			texture.LoadImage(data);
+			Progress.TextureLoaded++;
```
pjramirez-ms: Check that things aren't getting double counted? Consider making the fields properties with private setters and add functions to increment the values. Then in your functions you can assert that the loaded count does not exceed the total count.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
rferrese: Done

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
rferrese: I added an assert to LoadSceneAsync as well as asserts to not double load ImageCache values

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
rferrese: See new commit for details

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
rferrese: Great idea.  I'll just have it return 0% for now since that will be basically the same.  For now, I'm saying that the progress is Loaded/Total including nodes AND textures.  Not everything takes the same amount of time, but hopefully this will be reasonable.

AdamMitchell-ms: Sounds good.  We can always improve the accuracy later if needed but this is a great improvement and is probably good enough.

Issue #411, opened 2019-04-17
Loading animation issue


PedroBGDev:
Hi,
I'm trying to load some glTF files with animations with the latest UnityGLTF version and I always get errors when try to load it. I used an glTF viewer to see if there is any problem with the animation schema, but it works, so I think that is an bug at the importer.

The error that Unity shows is this, and there seems to be something uninitialized:
`NullReferenceException: Object reference not set to an instance of an object
UnityGLTF.GLTFSceneImporter.BuildAnimationSamplers (GLTF.Schema.GLTFAnimation animation, System.Int32 animationId) (at Assets/GLTF/UnityGLTF/Scripts/GLTFSceneImporter.cs:792)
UnityGLTF.GLTFSceneImporter.ConstructClip (UnityEngine.Transform root, UnityEngine.GameObject[] nodes, System.Int32 animationId) (at Assets/GLTF/UnityGLTF/Scripts/GLTFSceneImporter.cs:885)
UnityGLTF.GLTFSceneImporter+<ConstructScene>d__66.MoveNext () (at Assets/GLTF/UnityGLTF/Scripts/GLTFSceneImporter.cs:1063)
--- End of stack trace from previous location where exception was thrown ---
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () (at <d7ac571ca2d04b2f981d0d886fa067cf>:0)
System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) (at <d7ac571ca2d04b2f981d0d886fa067cf>:0)
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) (at <d7ac571ca2d04b2f981d0d886fa067cf>:0)
System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) (at <d7ac571ca2d04b2f981d0d886fa067cf>:0)
System.Runtime.CompilerServices.TaskAwaiter.GetResult () (at <d7ac571ca2d04b2f981d0d886fa067cf>:0)
UnityGLTF.GLTFSceneImporter+<_LoadScene>d__55.MoveNext () (at Assets/GLTF/UnityGLTF/Scripts/GLTFSceneImporter.cs:555)
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #429, opened 2019-05-23
Fix a crash when loading some animations

AdamMitchell-ms:
Solves issue #411.

Fix several places where we would assume that a buffer or node had been loaded when it in fact had not.  Now we always ensure that they have loaded before proceeding.

BREAKING CHANGE - The signature of BuildAnimationSamplers(), ConstructClip(), and SetupBones() have all been modified.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -890,28 +890,24 @@ namespace UnityGLTF
 				var samplerDef = animation.Samplers[i];
 
 				// set up input accessors
-
-				var inputBuffer = samplerDef.Input.Value.BufferView.Value.Buffer;
-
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -890,7 +890,11 @@ protected virtual void BuildAnimationSamplers(GLTFAnimation animation, int anima
 				var samplerDef = animation.Samplers[i];
 
 				// set up input accessors
-				BufferCacheData bufferCacheData = _assetCache.BufferCache[samplerDef.Input.Value.BufferView.Value.Buffer.Id];
+
+				var inputBuffer = samplerDef.Input.Value.BufferView.Value.Buffer;
```
pjramirez-ms: Doesn't look like this new variable is used

AdamMitchell-ms: Good catch.  I've removed it now and changed it to use two separate variables for input and output bufferCacheData rather than reusing the same variable.

PR #431, opened 2019-05-28
moved GLTFUri trim inside if statement to only trim when path.combine is needed

### LINKED ISSUE #430: GLTFComponent path.combine fails runtime import from network share
When you import at run-time from a network share the below code meant to fix path.combine trims the starting \\\\ and causes invalid paths.

example:
"\\\\server\share\file" becomes "server\share\file"

GLTFComponent.cs Line 79
```
// Path.Combine treats paths that start with the separator character
// as absolute paths, ignoring the first path passed in. This removes
// that character to properly handle a filename written with it.
GLTFUri = GLTFUri.TrimStart(new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar });
```

I can't seem to think of how to accommodate both cases off the top of my head but I'll come back to this and make a pull request if I come up with something good enough

fagerbur:
Fixes #430

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFComponent.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFComponent.cs
@@ -74,14 +74,13 @@ namespace UnityGLTF
 			{
 				if (UseStream)
 				{
-					// Path.Combine treats paths that start with the separator character
-					// as absolute paths, ignoring the first path passed in. This removes
-					// that character to properly handle a filename written with it.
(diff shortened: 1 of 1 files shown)


AdamMitchell-ms:
@fagerbur ,

Thanks for the submission!  This looks good to me.

Have you tested all of these scenarios to ensure they all work?
- network share path
- StreamingAssets path
- http path

If so, then I'll go ahead and complete your PR.

fagerbur:
I've tested network share and streaming assets path. For http path, I thought the way it worked was you would set UseStream to false which would then use the web request loader instead of the file loader so it wouldn't ever come across the trim change using path.combine.

PR #441, opened 2019-06-04
Update sample list so it can view generated asset list

AdamMitchell-ms:
- Sample list view can now parse and display the models from gltf-Asset-Generator.
- Update to the latest gltf-Asset-Generator and gltf-Sample-Models.
- Create a custom GLTFComponent inspector that shows a list of animations.


AdamMitchell-ms:
Superseded by pull #442

PR #442, opened 2019-06-04
Improve animation interpolation plus more sample scene editor integrations

AdamMitchell-ms:
- Correctly handle the fact that cubic spline interpolated animations have 3 values per keyframe instead of one.
  - in tangent
  - data point
  - out tangent
- Get linear, step, and catmullrom interpolation almost working correctly.
  - When looping forward and backward, going from the last keyframe backwards to the previous one does not seem to interpolate correctly.  Not sure why just yet.
- Sample list view can now parse and display the models from gltf-Asset-Generator.
- Update to the latest gltf-Asset-Generator and gltf-Sample-Models.
- Create a custom GLTFComponent inspector that shows a list of animations.


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -949,25 +952,77 @@ protected virtual async Task BuildAnimationSamplers(GLTFAnimation animation, int
...
+				{
+					// For cubic spline, the output will contain 3 values per keyframe; inTangent, dataPoint, and outTangent.
+					// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#appendix-c-spline-interpolation
+
+					var cubicIndex = i * 3;
+					inTangents = getConvertedValues(output, cubicIndex);
+					values = getConvertedValues(output, cubicIndex + 1);
+					outTangents = getConvertedValues(output, cubicIndex + 2);
+				}
+				else
```
azakariaMSFT: suggest adding the full condition if(interpolationmode =keyframe ||  something else )  in this case to make it clear what type of interpolation that goes here

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -949,25 +952,77 @@ protected virtual async Task BuildAnimationSamplers(GLTFAnimation animation, int
...
 			{
-				// set interpolcation for each keyframe
-				SetCurveMode(keyframes[ci], mode);
 				// copy all key frames data to animation curve and add it to the clip
 				AnimationCurve curve = new AnimationCurve();
 				curve.keys = keyframes[ci];
+
+				// For cubic spline interpolation, the inTangents and outTangents are already explicitly defined.
+				// For the rest, set them appropriately.
+				if (mode != InterpolationType.CUBICSPLINE)
```
azakariaMSFT: same, suggest being explicit with the type you need this code for. make it easy to add new animation types to the standard without causing going into unexpected paths

UnityGLTF/Assets/UnityGLTF/Scripts/Tests/Editor/SampleModelListInspector.cs
```diff
@@ -6,106 +6,178 @@
...
+
+		private void DrawModel(SampleModel model)
+		{
+			//bool isFocused = model == currentModel;
+
+			using (var horizontal = new EditorGUILayout.HorizontalScope())
+			{
+				if (model.Expanded)
+				{
+					//GUIStyle style = new GUIStyle(GUI.skin.label);
```
azakariaMSFT: remove commented code

UnityGLTF/Assets/UnityGLTF/Scripts/Tests/Editor/SampleModelListParser.cs
```diff
@@ -1,155 +1,334 @@
...
+				|| reader.Value.ToString().ToLowerInvariant() != expectedString)
+			{
+				throw new Exception($"Failed to parse {expectedString}");
+			}
+			reader.Read();
+		}
 
-		return variants;
-	}
+		//private static string ParseStringValue(JsonReader reader)
```
pjramirez-ms: Commented out code

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
AdamMitchell-ms: This doesn't seem like the kind of thing that is likely to change in the standard.  I could see it changing via an extension, but then the person adding support for the standard would naturally end up fixing things here.

PR #444, opened 2019-06-07
Remove use of UnityEditor namespace

AdamMitchell-ms:
UnityEditor namespace is not available when you build a player.  Implement an equivalent function as the helper that is unavailable.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -995,53 +995,53 @@ namespace UnityGLTF
 				{
 					for (var i = 0; i < keyframes[ci].Length; i++)
 					{
-                        SetTangentMode(curve, i, mode);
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -996,34 +995,55 @@ protected virtual async Task BuildAnimationSamplers(GLTFAnimation animation, int
 				{
 					for (var i = 0; i < keyframes[ci].Length; i++)
 					{
-						var utilityMode = GetAnimationUtilityMode(mode);
-
-						AnimationUtility.SetKeyLeftTangentMode(curve, i, utilityMode);
-						AnimationUtility.SetKeyRightTangentMode(curve, i, utilityMode);
-					}
-				}
+                        SetTangentMode(curve, i, mode);
```
AdamMitchell-ms: Tabulation seems wrong.  I'll clean it up.

PR #448, opened 2019-06-12
Fix stack overflow that could occur for some models

AdamMitchell-ms:
For some models, there are circular references where a node points to a mesh which points to a skinned animation, which points back to the original node.  This caused infinite recursion resulting in a stack overflow.

The fix is to load the subnodes prior to loading the mesh data.  There might be a better fix where all nodes are loaded, then all meshes are loaded.  But that was a much more substantial fix, so going with this one for now.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -1321,7 +1321,6 @@ namespace UnityGLTF
 					for (int i = 0; i < lodsextension.MeshIds.Count; i++)
 					{
 						int lodNodeId = lodsextension.MeshIds[i];
-						Debug.Log("Loading lod extension");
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1344,11 +1321,11 @@ protected virtual async Task ConstructNode(Node node, int nodeIndex, Cancellatio
 					for (int i = 0; i < lodsextension.MeshIds.Count; i++)
 					{
 						int lodNodeId = lodsextension.MeshIds[i];
-						await ConstructNode(_gltfRoot.Nodes[lodNodeId], lodNodeId, cancellationToken);
-						int lodIndex = i + 1;
-						GameObject lodNodeObj = _assetCache.NodeCache[lodNodeId];
+						Debug.Log("Loading lod extension");
```
AdamMitchell-ms:
```suggestion
```

Issue #432, opened 2019-05-28
Out Of Memory Handling


rferrese:
On iOS (though possibly on other platforms) out of memory exceptions caused by Texture2D.LoadImage cause the app to crash.  This is because the exception happens on gfx thread, and isn't caught by a try/catch around LoadImage(...)  Ideally, running out of memory would cause the load to fail, discard all loaded data, and continue app execution normally.

AdamMitchell-ms:
Fixed

PR #454, opened 2019-06-20
Handle low memory situations and change API of Importer/Exporter to take in option classes

AdamMitchell-ms:
- Handle low memory situations when loading a scene.  Fixes issue #432.
   - Under memory pressure, calls to Texture2D.LoadImage() can cause the app to crash in an un-catchable way.
   - On iOS, Android, and non-desktop Windows (Hololens, etc.), the OS will notify an app if it is using too much memory before it actually runs out.
   - We now listen to low memory events during loading.  If it fires, we remove the partially loaded model and throw an OutOfMemoryException.  Callers can opt out of this functionality using the ImportOptions if they wish.
- Change the API of GLTFSceneImporter and GLTFSceneExporter to take in option objects.
   - Old API is marked as obsolete, but will still work for now.
   - The plan is to remove the old constructors in a future change.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
@@ -11,15 +11,15 @@ using WrapMode = GLTF.Schema.WrapMode;
 
 namespace UnityGLTF
 {
+	public class ExportOptions
+	{
+		public GLTFSceneExporter.RetrieveTexturePathDelegate TexturePathRetriever = (texture) => texture.name;
+		public bool ExportInactivePrimitives = true;
+	}
+
 	public class GLTFSceneExporter
 	{
 		public delegate string RetrieveTexturePathDelegate(Texture texture);
-
-		public class ExportOptions
-		{
-			public RetrieveTexturePathDelegate TexturePathRetriever = (texture) => texture.name;
-			public bool ExportInactivePrimitives = true;
-		}
 
 		private enum IMAGETYPE
 		{
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
@@ -15,6 +15,12 @@ namespace UnityGLTF
 	{
 		public delegate string RetrieveTexturePathDelegate(Texture texture);
 
+		public class ExportOptions
+		{
+			public RetrieveTexturePathDelegate TexturePathRetriever = (texture) => texture.name;
+			public bool ExportInactivePrimitives = true;
+		}
+
 		private enum IMAGETYPE
 		{
 			RGB,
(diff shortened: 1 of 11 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -2196,6 +2272,19 @@ protected virtual ExtTextureTransformExtension GetTextureTransform(TextureInfo d
 			else return null;
 		}
 
+		protected async Task YieldOnTimeoutAndCheckMemory()
+		{
+			if (_options.CheckForLowMemory)
```
pjramirez-ms: Nit: 'Check' is kind of vague. Consider renaming to ThrowOnLowMemory to make it clear what the flag does

UnityGLTF/Assets/UnityGLTF/Scripts/MemoryChecker.cs
```diff
@@ -0,0 +1,53 @@
...
+		{
+			outOfMemory = true;
+		}
+
+#if WINDOWS_UWP
+		private void MemoryManager_AppMemoryUsageIncreased(object sender, object e)
+		{
+			if (MemoryManager.AppMemoryUsageLevel == AppMemoryUsageLevel.OverLimit)
+			{
+				outOfMemory = true;
```
pjramirez-ms: Is there a way to check memory again and set this to false if memory gets cleaned up?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -175,45 +183,67 @@ protected struct GLBStream
...
 			_gltfRoot = rootNode;
-			_loader = externalDataLoader;
+
 			if (gltfStream != null)
 			{
 				_gltfStream = new GLBStream { Stream = gltfStream, StartPosition = gltfStream.Position };
 			}
 		}
 
+		[Obsolete]
```
pjramirez-ms: Nit: add obselete comment here too?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneExporter.cs
```diff
@@ -82,9 +88,19 @@ protected struct PrimKey
 		/// Create a GLTFExporter that exports out a transform
 		/// </summary>
 		/// <param name="rootTransforms">Root transform of object to export</param>
-		public GLTFSceneExporter(Transform[] rootTransforms, RetrieveTexturePathDelegate retrieveTexturePathDelegate)
+		[Obsolete]
```
admachina: Can you put a comment on this obsolete tag also?

UnityGLTF/Assets/UnityGLTF/Scripts/MemoryChecker.cs
AdamMitchell-ms: Unfortunately, no.  It is not really possible.  The underlying events from Android and Windows _do_ tell you when they are no longer low on memory.  But it doesn't appear that iOS exposes an event like that.  So Unity does not expose any event for it.  The expectation is that the app must reduce its memory usage when that event fires.  If your process is not terminated, then I guess that means that you're no longer low on memory?

I considered setting outOfMemory to false when throwing, assuming that it will go down and then could go up again and then we'd get the event again, but I don't know if that would work well.

So I decided to put the comment at the top explaining the expected usage.  If it throws, then the caller should clear some memory and create a new MemoryChecker.

Issue #451, opened 2019-06-14
Does not work at all with Android or IOS


anupamdas2012:
tested on both Android and Ios with unity 2019.1.4f1. Tried the web based and non web based (UWP) examples, nothing works on either devices. Assuming mobile devices is not supported currently?

AdamMitchell-ms:
When you say that nothing works, can you be more specific?  UnityGLTF should work for iOS and Android.  If it's broken in some way, then we should get it fixed.

marcspraragen:
I'm also having trouble running on Android. I tried builds with one scene, all scenes, and my own scene. The apk runs but just displays a blank game screen (skybox and floor). No way to navigate between scenes. 

Also, no console errors, just warnings:

>Script attached to 'GLTF' in scene 'Assets/UnityGLTF/Examples/StreamTestScene.unity' is missing or no valid script is attached.
UnityEditor.BuildPlayerWindow:BuildPlayerAndRun()

>Assets/UnityGLTF/Examples/OrbitCameraController.cs(22,22): warning CS0109: The member `UnityGLTF.Examples.OrbitCameraController.camera' does not hide an inherited member. The new keyword is not required

antoninklopp:
The problem is that File.ReadAllBytes() is not working for Assets on my side from my side. 
I get a FileNotFoundException on Android and Oculus

jingwenhz:
I also see only the skybox and floor when building on Android. adb logcat output shows:

`Unity   : ArgumentOutOfRangeException: Length cannot be less than zero.`
`Unity   : Parameter name: length`
`Unity   :   at System.String.Substring (System.Int32 startIndex, System.Int32 length) [0x0004a] in <1f0c1ef1ad524c38bbc5536809c46b48>:0 `
`Unity   :   at UnityGLTF.Examples.WebServerComponent.Start () [0x00007] in /Users/.../UnityGLTF/UnityGLTF/Assets/UnityGLTF/Examples/WebServerComponent.cs:12 
`

The line in question in WebServerComponent.cs is 
`var wwwPath = appPath.Substring(0, appPath.LastIndexOf("Assets")) + "www";`

I have no problems running/importing models in the Unity editor. Do we need to take additional steps to properly run scenes on mobile?

antoninklopp:
Hi, 

After a few tries on Unity Android, I managed to compile an Apk and got the plugin to work. 
You need to do a few things : 
[...]

PR #460, opened 2019-06-26
Update README.md

antoninklopp:
Updated README after Issue #451 .

A good idea could also be to add the shader in the [GraphicsSettings](https://github.com/KhronosGroup/UnityGLTF/blob/master/UnityGLTF/ProjectSettings/GraphicsSettings.asset) from the sample project and explain everything in a Build.MD since some of the README.md is directly related to build explanations.


marcspraragen:
Hi-- is there an update of readme.md I could look at?

antoninklopp:
Hi,

I will push a new update because I need to tell that the build on android should include both shaders

antoninklopp:
Hi @marcspraragen , 

I just pushed the update

hybridherbst:
Closed – outdated. The Readme got rewritten for UnityGLTF 2.x.

PR #464, opened 2019-07-12
fix chunkOffset bug

verybigzhouhai:
When the startPosition is not equal to zero, the chunkOffset should add the startPosition.


verybigzhouhai:
I recently used this project to load the 3dtiles file. When I passed in the whole stream of b3dm, I needed to set the startPosition not to be zero. When the chunk offset is added, otherwise I would get the wrong chunk length value.

verybigzhouhai:
need change account

PR #465, opened 2019-07-12
fix chunkOffset bug

verybigzhouhai:
When the startPosition is not equal to zero, the chunkOffset should add the startPosition

### DIFF
--- a/GLTFSerialization/GLTFSerialization/GLTFParser.cs
+++ b/GLTFSerialization/GLTFSerialization/GLTFParser.cs
@@ -60,7 +60,7 @@ namespace GLTF
 		{
 			stream.Position = startPosition + 4;	 // start after magic number chunk
 			GLBHeader header = ParseGLBHeader(stream);
-			uint chunkOffset = 12;   // sizeof(GLBHeader) + magic number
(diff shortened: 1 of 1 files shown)


verybigzhouhai:
I recently used this project to load the 3dtiles file. When I passed in the whole stream of b3dm, I needed to set the startPosition not to be zero. When the chunk offset is added, otherwise I would get the wrong chunk length value.

AdamMitchell-ms:
Hi @verybigzhouhai ,
I just approved PR #458.  Can you merge the latest master into your branch and verify that it still works correctly?  They sound like similar problems and I want to make sure that your changes are compatible.

verybigzhouhai:
Hi, @AdamMitchell-ms Thanks for your reply, it seems that this is not the same problem, because this function sets the startPosition parameter, then this parameter will have a situation that is not equal to 0, and this should be allowed if the stream data we pass in is not A standard glb, like the b3dm file I used, needs to include this starting position when calculating this offset, because the standard glb header is 12, which looks more like a logic error in the function itself. Because all calculations should include this parameter because it is already used in the first line of the function.

verybigzhouhai:
@AdamMitchell-ms Please see if I judged correctly.
Thank you very much

AdamMitchell-ms:
Ok.  Sounds good.

PR #477, opened 2019-08-07
Add API for loading meshes on their own

stevenvergenz:


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -416,6 +426,63 @@ public virtual async Task<Material> LoadMaterialAsync(int materialIndex)
...
+                {
+                    if (_isRunning)
+                    {
+                        throw new GLTFLoadException("Cannot CreateMesh while GLTFSceneImporter is already running");
+                    }
+
+                    _isRunning = true;
+                }
+
+				if (_options.ThrowOnLowMemory)
```
AdamMitchell-ms: Tabs here.  Should be spaces.

AdamMitchell-ms: A lot of this boilerplate is shared in common with LoadSceneAsync, LoadMaterialAsync, and LoadNodeAsync.  Consider wrapping it in a Setup() function that lives next to Cleanup().

Specifically, constructing _memoryChecker and _assetCache, and calling LoadJson().

stevenvergenz: Actually the whole file is tabs except for this function. Updated this to be consistent.

stevenvergenz: Good call! Done.

AdamMitchell-ms: Oh yeah, you're right.  I had it backwards.

PR #479, opened 2019-08-18
Optimized animation importing in GLTFSceneImporter by storing the keyframes in a variable

michielbos:
It seems that importing of animations takes a long time for some models. After a bit of digging, it seemed that the main cause is that the AnimationCurve.keys array is being accessed in multiple places, which causes it to make multiple copies of itself (see [AnimationCurve.keys](https://docs.unity3d.com/ScriptReference/AnimationCurve-keys.html)).

Storing the keyframes in a variable instead of requesting new copies seems to reduce the loading time from 2 seconds per model to about 150 milliseconds per model for my project.

I hope this pull request helps. It's the first time I'm submitting a pull request to a public repo so I hope I did everything right. ^^

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -1012,8 +1012,7 @@ namespace UnityGLTF
 			for (var ci = 0; ci < channelCount; ++ci)
 			{
 				// copy all key frames data to animation curve and add it to the clip
-				AnimationCurve curve = new AnimationCurve();
-				curve.keys = keyframes[ci];
+				AnimationCurve curve = new AnimationCurve(keyframes[ci]);
 
 				// For cubic spline interpolation, the inTangents and outTangents are already explicitly defined.
 				// For the rest, set them appropriately.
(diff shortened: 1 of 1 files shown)


blgrossMS:
@AdamMitchell-ms  can you take a look?

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1012,25 +1012,23 @@ protected virtual async Task BuildAnimationSamplers(GLTFAnimation animation, int
...
 					}
 				}
 				clip.SetCurve(relativePath, curveType, propertyNames[ci], curve);
 			}
 		}
 
-		private static void SetTangentMode(AnimationCurve curve, int keyframeIndex, InterpolationType interpolation)
-		{
-			var key = curve.keys[keyframeIndex];
+		private static void SetTangentMode(AnimationCurve curve, Keyframe[] keyframes, int keyframeIndex, InterpolationType interpolation) {
```
AdamMitchell-ms: put the open brace on the next line

michielbos:
@AdamMitchell-ms Thanks for the review. I've fixed the curly bracket.

PR #490, opened 2019-09-17
Throwing the exception from the SetUpLoad to the caller method, so that

SaiKanth007:
Throwing the exception from the SetUpLoad to the caller method, so that it can be used later on


SaiKanth007:
Closing this as we have raised an other [PR ](https://github.com/KhronosGroup/UnityGLTF/pull/494)for the same changes and signed CLA as well,

PR #492, opened 2019-09-19
Throwing the exception after cleanup from SetUpLoad

tgjaikish:
Throwing the exception after cleanup from SetUpLoad to the caller method, so that it can be used later on


tgjaikish:
Change committed through a different PR

PR #495, opened 2019-09-23
Fix camel-casing on wraps

hexuallyactive:


AdamMitchell-ms:
Thanks for the submission @hexuallyactive

PR #497, opened 2019-10-02
Proper package structuring for Unity Package Manager

soraryu:
As Unity is moving more and more towards using Package Manager for everything, this project should really be properly using the suggested package structure.

This PR moves everything into the right places, but there's a couple things left to do:
- the package creation shell scripts should rename the Samples folder to Samples~ (these can be accessesed by users from within package manager as proper Samples then)
- the Tests should be converted to actually use Unity TestRunner (and show up there)
- a CHANGELOG.md should be added

This is how it currently looks in Packages now (note the missing rename to Samples~ so they don't show by default) compared to the USDZ package:
image


blgrossMS:
This looks great to me, @AdamMitchell-ms  can you take a look?

soraryu:
I dealt with the conflicts (actually GitHub was just not smart enough, git just was able to cleanly rebase as the only problem was that files that I moved had been edited in newer commits).

Please note that ideally right after merging this in the following should be done:
- Plugins should be compiled into the "Runtime/Plugins" folder, not into "Plugins" as right now
- Samples should be renamed to Samples~ when the UPM package export shell scripts run

AdamMitchell-ms:
Hi @soraryu,

I just found out that this file restructuring caused the play mode tests to stop appearing.  I know it was a while ago, but can you take a look?  I've created issues #526 to track it.

AdamMitchell-ms:
Ok, actually I was able to fix it pretty simply.  I have a PR out to fix it now.

soraryu:
OK :) Should just be moving it to the right "Testing" folder and making sure the asmdefs are set up for testing.

PR #498, opened 2019-10-02
Remove O(N*M) asset reimports through importer.SaveAndReimport - 200x import speed improvement for test file

hybridherbst:
Currently, putting GLB files with a couple of textures into the project folder results in minute-long waits as the importer script has an inner loop calling TextureImporter.SaveAndReimport.

For a simple object (attached) with ~22 objects with about a texture each (some share textures), this results in crazy import times:

This PR replaces those by a single call to AssetDatabase.SaveAssets and a subsequent AssetDatabase.Refresh, resulting in sub-second import times (down from > 200s before).
A 20.000% speed increase!

Additionally, this PR adds a try/catch around the AssetDatabase context add as that fails imports while it should actually only produce warnings.

This is the file to test this PR with:
[comp-20190923-174143.zip](https://github.com/KhronosGroup/UnityGLTF/files/3681972/comp-20190923-174143.zip)

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Editor/Scripts/GLTFImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Editor/Scripts/GLTFImporter.cs
@@ -279,8 +279,6 @@ namespace UnityGLTF
                                                 // Force disable sprite mode, even for 2D projects
                                                 importer.textureType = TextureImporterType.Default;
                                             }
-
-                                            //importer.SaveAndReimport();
                                         }
                                         else
                                         {
--- a/UnityGLTF/Assets/UnityGLTF/Editor/Scripts/GLTFImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Editor/Scripts/GLTFImporter.cs
@@ -211,6 +211,8 @@ namespace UnityGLTF
                         }
                     }
 
+					AssetDatabase.Refresh();
+
                     // Save materials as separate assets and rewrite refs
                     if (materials.Length > 0)
                     {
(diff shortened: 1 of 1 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
```diff
@@ -278,7 +280,7 @@ public override void OnImportAsset(AssetImportContext ctx)
                                                 importer.textureType = TextureImporterType.Default;
                                             }
 
-                                            importer.SaveAndReimport();
+                                            //importer.SaveAndReimport();
```
AdamMitchell-ms: Rather than commenting, just remove the line.

soraryu:
I removed the commented line, thanks for the feedback.

Interesting that you don't see the slow import times.
Here's exact repro steps (on Windows 10):
- be on master
- make sure to build GLTFSerialization\GLTFSerialization from GLTFSerialization\GLTFSerialization.sln
- open the UnityGLTF project in 2018.3.14f1
- unzip the file attached on the PR
- drag and drop the file comp-20190923-174143.glb into Unity
**ACTUAL (on master)**
- see - and wait, and wait, and wait (~3 minutes on a high-end machine)
- note that after the long wait the model partially imports but some objects are missing; console logs error messages `InvalidOperationException: The object Saturn 1(Clone) was already added to the context.` and `Asset import failed, "Assets/comp-20190923-174143.glb" > InvalidOperationException: The object Saturn 1(Clone) was already added to the context.`.
**EXPECTED (with this PR)**
- import goes pretty quickly (1-5s depending on Unity version)
- import works, no error message

I was able to reproduce this in these versions I had at hand:
- Unity 2018.3.14f1
- Unity 2019.1.8f1
- Unity 2019.2.8f1
- Unity 2019.3.0b4
- Unity 2020.1.0a3

soraryu:
@AdamMitchell-ms did you have a chance to look at this again?

PR #499, opened 2019-10-02
Better handling of attribute errors

soraryu:
Instead of throwing when finding unrecognized attributes, this PR makes sure all attributes are collected and an exception is thrown after all are known. This allows to actually catch the exception and handle it per object instead of just failing.

Note that I did this because TEXCOORD_4 throws exceptions (it's not available in the attribute list) - not sure if this is a bug in itself.

A file to test this with is attached. I'm not entirely sure why it even has TEXCOORD_4, but with this PR the object imports just fine.
[wooden_eagle.zip](https://github.com/KhronosGroup/UnityGLTF/files/3682255/wooden_eagle.zip)

### DIFF
--- a/GLTFSerialization/GLTFSerialization/Exceptions.cs
+++ b/GLTFSerialization/GLTFSerialization/Exceptions.cs
@@ -26,15 +26,15 @@ namespace GLTF
 #endif
 	}
 
-    public class GLTFLoadException : Exception
-    {
-        public GLTFLoadException() : base() { }
-        public GLTFLoadException(string message) : base(message) { }
-        public GLTFLoadException(string message, Exception inner) : base(message, inner) { }
+	public class GLTFLoadException : Exception
+	{
+		public GLTFLoadException() : base() { }
+		public GLTFLoadException(string message) : base(message) { }
+		public GLTFLoadException(string message, Exception inner) : base(message, inner) { }
 #if !WINDOWS_UWP
-        protected GLTFLoadException(System.Runtime.Serialization.SerializationInfo info,
-            System.Runtime.Serialization.StreamingContext context)
-        { }
+		protected GLTFLoadException(System.Runtime.Serialization.SerializationInfo info,
+			System.Runtime.Serialization.StreamingContext context)
+		{ }
 #endif
-    }
+	}
 }
--- a/GLTFSerialization/GLTFSerialization/Exceptions.cs
+++ b/GLTFSerialization/GLTFSerialization/Exceptions.cs
@@ -25,4 +25,16 @@ namespace GLTF
 		{ }
 #endif
 	}
+
+    public class GLTFLoadException : Exception
+    {
+        public GLTFLoadException() : base() { }
+        public GLTFLoadException(string message) : base(message) { }
+        public GLTFLoadException(string message, Exception inner) : base(message, inner) { }
+#if !WINDOWS_UWP
+        protected GLTFLoadException(System.Runtime.Serialization.SerializationInfo info,
+            System.Runtime.Serialization.StreamingContext context)
+        { }
+#endif
+    }
 }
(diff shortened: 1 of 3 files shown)


GLTFSerialization/GLTFSerialization/GLTFHelpers.cs
```diff
@@ -260,11 +261,21 @@ public static void BuildMeshAttributes(ref Dictionary<string, AttributeAccessor>
...
 				attributeAccessor.AccessorContent = resultArray;
 			}
+
+			foreach (var attrib in unrecognizedAttributes) { 
+				attributes.Remove(attrib);
+			}
+
+			// TODO This should be a warning. Unrecognized attributes (e.g. TEXCOORD_4) should not cause full exceptions.
+			if (unrecognizedAttributes.Count > 0) {
+				throw new System.Exception($"Unrecognized mesh attributes [{string.Join(", ", unrecognizedAttributes.ToArray())}]");
```
AdamMitchell-ms: >System.Exception [](start = 14, length = 16)

Not exactly your change, but this should probably be a GLTFLoadException.

UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -844,8 +844,11 @@ protected virtual async Task ConstructPrimitiveAttributes(MeshPrimitive primitiv
 					Offset = (uint)bufferData.ChunkOffset
 				};
 			}
-
-			GLTFHelpers.BuildMeshAttributes(ref attributeAccessors);
+			try { 
+				GLTFHelpers.BuildMeshAttributes(ref attributeAccessors);
+			} catch (System.Exception e) {
```
AdamMitchell-ms: >System.Exception [](start = 12, length = 16)

As a general rule, you should never catch System.Exception and swallow it.  There are million things it could be and you have no way to know if it's reasonable to continue on.

You can catch the specific one that I suggested you change it to (GLTFLoadException).

GLTFSerialization/GLTFSerialization/GLTFHelpers.cs
soraryu: Ah, good, I was looking for the right one but didn't find it on first glance. I'll change it, of course don't want to catch all.

soraryu: OK, think I misunderstood you but I now added a GLTFLoadException for this and future usecases.

soraryu:
OK, I added that new exception type (following the pattern for others in Exceptions.cs) and used it to catch those in this PR.

GLTFSerialization/GLTFSerialization/Exceptions.cs
```diff
@@ -25,4 +25,16 @@ public class GLTFParseException : Exception
 		{ }
 #endif
 	}
+
+    public class GLTFLoadException : Exception
```
AdamMitchell-ms: >     [](start = 0, length = 4)

nit: should be tabs.  I know we're inconsistent in some files, but the .editorconfig says tabs.  So everything should be tabs.

AdamMitchell-ms:
Ok, I just merged in your big change that moved files, so now there is a conflict.  It should be easy for you to fix up (plus my one new comment about tabs) and then we can merge it in.

soraryu:
Yup, fixed

EricBeetsOfficial:
> Instead of throwing when finding unrecognized attributes, this PR makes sure all attributes are collected and an exception is thrown after all are known. This allows to actually catch the exception and handle it per object instead of just failing.
> 
> Note that I did this because TEXCOORD_4 throws exceptions (it's not available in the attribute list) - not sure if this is a bug in itself.
> 
> A file to test this with is attached. I'm not entirely sure why it even has TEXCOORD_4, but with this PR the object imports just fine.
> [wooden_eagle.zip](https://github.com/KhronosGroup/UnityGLTF/files/3682255/wooden_eagle.zip)


But the model loading failed cause in GLTFSceneImporter.cs:881 the `GLTFLoadException` is from UnityGLTF.GLTFLoadException  (Exceptions.cs:35) and not from the GLTF.GLTFLoadException of the GLTFSerialization part (Exceptions.cs: 29).
Why is GLTFLoadException redefined in UnityGLTF ? Shouldn't you delete the class from UnityGLTF (and keep it in the GLTFSerialization part) ?

soraryu:
I think this might have been an oversight on my end. I searched for the exception type but didn't find an existing one. Feel free to make a PR to correct this, thank you!

PR #500, opened 2019-10-03
Adding ability to get vertex count and triangle count from model load

blgrossMS:
- Vertex count and triangle count not gotten after load from Statistics on GLTFSceneImporter
- TriangleCount only accurate for meshes that have indices and a triangle topology

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
@@ -1634,7 +1634,7 @@ namespace UnityGLTF
 				var vertCount = primitive.Attributes[SemanticProperties.POSITION].Value.Count;
 				vertOffset += (int)vertCount;
 
-				if (unityData.Topology[i] == MeshTopology.Triangles)
+				if (unityData.Topology[i] == MeshTopology.Triangles && primitive.Indices != null && primitive.Indices.Value != null)
 				{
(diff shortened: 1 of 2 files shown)


UnityGLTF/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1620,8 +1633,14 @@ protected virtual async Task ConstructMesh(GLTFMesh mesh, int meshIndex, Cancell
 
 				var vertCount = primitive.Attributes[SemanticProperties.POSITION].Value.Count;
 				vertOffset += (int)vertCount;
+
+				if (unityData.Topology[i] == MeshTopology.Triangles && primitive.Indices != null && primitive.Indices.Value != null)
+				{
+					Statistics.TriangleCount += primitive.Indices.Value.Count / 3;
```
RickyRoeslerMS: Would it be better to expose this out more generically so it could count whatever topological unit the mesh uses? The number of triangles or quads or points or lines?

blgrossMS: Good call out. I'll implement this for Quad and Triangle and also put topology in the stats

blgrossMS: Actually I'll leave it like this for now. Technically different submeshes can have different topologies, which would mean a counter for topo type is needed. The asset cache can currently be queried after if the user is interested in getting data per topo

PR #506, opened 2019-10-07
Optimize animation import

ashikns:
Improves upon three things:

1. Do tangent calculation operations on keyframe cached array rather than AnimationCurve. This is the most important change. Unity profiler shows that accessing AnimationCurve.keys is a costly operation.
2. Optimizes branching by moving if..else outside loop. The compiler probably already does this but it's nice to do it ourselves.
3. Fixes a bug when two incoming animation keys have the same time value. I don't know if this is valid per gltf specs but I faced this using some models downloaded from Sketchfab.


ashikns:
Just saw that there's another PR that includes #1 from this: #479

blgrossMS:
Thank you for contributing the change! @AdamMitchell-ms  can you review?

Issue #502, opened 2019-10-07
UnityGLTF Build Failing


blgrossMS:
The UnityGLTF Build is failing making it hard to determine quality of changes in PR

mvgalle:
Open GLTFSerialization.sln in Visual Studio, then do the following:
First: Build > Clean Solution. 
Second: Build > Build Solution.

mhochk:
Finished getting the build running the rest of the way with #532

PR #517, opened 2019-11-07
Build improvements take 2

mhochk:
Brings the Unity version into alignment (so the build an local versions are using the same thing) on the latest version of 2017 supported by the current install script.
Moved a few files around that were fallout from the original restructuring to have a Runtime folder.


mhochk:
AppVeyor build fixed, partially addressing #502.  The Travis-CI build failure appears to be rooted in the fact that on newer versions of Unity it properly checks for an activated license (even on Mac) so to use a newer version of Unity and have the Unity portion of the build work will require a greater investigation.
https://forum.unity.com/threads/package-manager-failed-to-resolve-packages-packages-directory-does-not-exist.522782/

PR #520, opened 2019-11-14
Update GLTFImporter.cs

MxShepherd:
Importing a model that reuses an instance of a mesh several times will throw an InvalidOperationException. This renders the asset invalid (no import settings show up). Skipping this error with try/catch leaves no errors.


MxShepherd:
CI check failed due to "Aborting batchmode due to failure: Asset not found: Assets/UnityGLTF".
But why?...

hybridherbst:
Closed – outdated. Importing this type of file does not throw an exception anymore.

PR #530, opened 2019-12-12
Mark asmdef as used for Testing

soraryu:
For some reason the Tests folder asmdef wasn't marked as test assembly for me; this results in compiler errors. This PR fixes that.


mhochk:
What compile errors was this generating for you?  I'd love to get the travis build validating that this kind of issue is not broken in the future if it's reasonable, but am not seeing any issue in my local builds.

soraryu:
It happens immediately when referencing the UnityGLTF/Assets/UnityGLTF folder (where the package.json resides) as package in another project, basically what we always do for local development where other projects need a package.

image
(this is in 2018.4.13f1)

(It will not happen when actually exporting a package - without the tests - but in our case it breaks when referencing).

This PR fixes that by setting the right options in the asmdef to mark it as testing asmdef:
image

Issue #503, opened 2019-10-07
Update UnityGLTF to 2018.4


blgrossMS:
UnityGLTF is currently on LTS for 2017.4, the main repo should be updated to 2018.4.

PR #535, opened 2019-12-19
Upgrade to Unity 2018

mhochk:
Updated to Unity 2018 (Issue #503)
- Updated documentation and actual project to 2018.4.14f1
- Overhauled the Travis build to use a docker image of a matching version of Unity.  (Includes running the playmode and editmode tests)
- Updated the skybox to individual images, since the .dds appears unsupported in 2018 (it was never *officially* supported)
- Updated lighting and unit tests, bringing them even closer to the provided sample images (though still quite a ways off from being able to image compare against them instead of our own _EXPECTED images).

Changes to Main and TestScenePrefab are primarily re-serialization from the Unity upgrade.  Practically the changes were adjusting the lighting and fixing the Model List to point at the loader.


scripts/before_script.sh
```diff
@@ -0,0 +1,38 @@
...
+mkdir -p /root/.cache/unity3d
+mkdir -p /root/.local/share/unity3d/Unity/
+set +x
+
+UPPERCASE_BUILD_TARGET=${BUILD_TARGET^^};
+
+if [ "$UPPERCASE_BUILD_TARGET" = "ANDROID" ]
+then
+    if [ -n $ANDROID_KEYSTORE_BASE64 ]
+	then
```
AdamMitchell-ms: >	 [](start = 0, length = 1)

nit: one tab in the file

mhochk: 🤦

Issue #540, opened 2019-12-26
Conversion error with Pbr shaders (Unity 2019, Android)


marcspraragen:
[   I'm using Unity 2019.3.0f3, 
UnityGLTF package manifest has 
	"name": "org.khronos.unitygltf",
	"displayName": "Khronos GLTF Loader",
	"version": "1.0.0",
	"unity": "2018.3",

I'm building for Android 7 (v24), using .NET 4.x API, mono scripting background. Testing on a Samsung 8+, though I haven't got it to build yet.  I have also saved the generated shader variant instructions to an asset as per previous issues here. ]

Before I do any script surgery (with very limited knowledge of these files) or revert to Unity 2018.3, I thought I'd check if anyone here has encountered this issue. I've attached the error log and the possible culprit VertexColor.cginc Thanks for your time.

--Marc

Error log: 
[GLTFconversionErrors.txt](https://github.com/KhronosGroup/UnityGLTF/files/4002784/GLTFconversionErrors.txt)

Here is the .cginc file that the errors are all blaming: 
[VertexColor.cginc.txt](https://github.com/KhronosGroup/UnityGLTF/files/4002786/VertexColor.cginc.txt)

marcspraragen:
Anybody at Khronos? This is a showstopper...

marcspraragen:
I fixed it. (desperation)

In shaders/VertexColor.cginc, change the class of "WrappedVertexOutputForwardBase.innerValue": 

struct WrappedVertexOutputForwardBase {
    //VertexOutputForwardBase innerValue;
	VertexOutputBaseSimple innerValue;
    fixed4 color : COLOR;
};

This ALSO seemed to fix #125 (UNITY_STANDARD_SIMPLE already defined).

PLEASE put this in readme, or just fix the master branch!

AdamMitchell-ms:
Hey @marcspraragen, can you create a pull request with your change?  @mhochk or I can review it and get it into master.

marcspraragen:
will do, and thanks :)

marcspraragen:
Request set. Also, a note: during initial build for android, variant compilation will throw "not found" errors for VertexOutputBaseSimple vs. the two shaders, but following builds work fine.

mhochk:
Fixed with #547

PR #213, opened 2018-07-30
Add UnityGLTFShaderVariants for issue: https://github.com/KhronosGrou…

Vytek:
…p/UnityGLTF/issues/206

See: https://github.com/KhronosGroup/UnityGLTF/issues/206

This include: UnityGLTFShaderVariant in UnityGLTF/Assets/Resources/ but also in GraphicsSettings.

Please check all ok!

Issue #125, opened 2018-02-10
Does not build for Android Platform Target


StephenHodgson:
Unity 2017.2
Android Build target

```
Shader error in 'GLTF/GLTFStandard': '' : 'UNITY_STANDARD_SIMPLE' already defined at /Program Files/Unity 2017.2/Editor/Data/CGIncludes/UnityStandardCoreForward.cginc(6) (on gles)
Error building Player: Shader error in 'GLTF/GLTFStandard': '' : 'UNITY_STANDARD_SIMPLE' already defined at /Program Files/Unity 2017.2/Editor/Data/CGIncludes/UnityStandardCoreForward.cginc(6) (on gles)
```

stevenvergenz:
Make sure you're using the latest version of UnityGLTF. That shader was replaced a few weeks ago.

keveleigh:
I'm seeing the same error on Android with the new shaders, tested with Unity 5.6, 2017.1, and 2017.2.

```
Shader error in 'GLTF/PbrMetallicRoughness': '' : 'UNITY_STANDARD_SIMPLE' already defined at /Program Files/Unity/Editor/Data/CGIncludes/UnityStandardCoreForward.cginc(6) (on gles)



Compiling Vertex program with DIRECTIONAL

Platform defines: UNITY_NO_DXT5nm UNITY_NO_RGBM UNITY_ENABLE_REFLECTION_BUFFERS UNITY_FRAMEBUFFER_FETCH_AVAILABLE UNITY_NO_CUBEMAP_ARRAY UNITY_NO_SCREENSPACE_SHADOWS UNITY_PBS_USE_BRDF3 UNITY_NO_FULL_STANDARD_SHADER SHADER_API_MOBILE UNITY_HARDWARE_TIER1 UNITY_COLORSPACE_GAMMA
```



```
Shader error in 'GLTF/PbrSpecularGlossiness': '' : 'UNITY_STANDARD_SIMPLE' already defined at /Program Files/Unity/Editor/Data/CGIncludes/UnityStandardCoreForward.cginc(6) (on gles)



Compiling Vertex program with DIRECTIONAL SHADOWS_SCREEN

[...]

PR #542, opened 2020-01-02
Update VertexColor.cginc to fix Android build

marcspraragen:
Addresses issues #125, #540. Builds on solutions presented by Vytek (#213) and antoninklopp (#451)

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Runtime/Shaders/VertexColor.cginc
+++ b/UnityGLTF/Assets/UnityGLTF/Runtime/Shaders/VertexColor.cginc
@@ -6,7 +6,8 @@
 #include "UnityStandardCoreForward.cginc"
 
 struct WrappedVertexOutputForwardBase {
-    VertexOutputForwardBase innerValue;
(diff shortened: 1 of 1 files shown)


marcspraragen:
Note: at first build, variant compilation will throw "identifier not found" errors for the two shaders, but following builds work fine.

mhochk:
There's actually a bigger issue here that I'll get started on a fix for shortly.  The root of the problem is the VertexColor.cginc code was made as a minimalistic wrapper of vertBase/fragBase, but doesn't include support for when UNITY_STANDARD_SIMPLE is set to true (which is apparently the case on Android builds).

This change currently let's you compile for Android, but shouldn't be working correctly for other targets (though I haven't tested to confirm this).  Also even though it compiles for Android, it's not using the shader intended by Unity for Android, so there's probably some kind of fallout there.

mhochk:
Got a fix together and put out PR #547

marcspraragen:
Oh wow thanks! I had noticed this too but didn't have time to set up a fix. I noticed it's up in master branch already.

PR #561, opened 2020-02-16
Export Unity animation clips and skeleton to GLTF

owilliamailliwo:
Add exporting Unity animation clips and skeleton to GLTF. We can export GLTF file with animation clips and skeleton throght "GLTF -> Export ***" menu now.

### DIFF
--- a/UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
@@ -1206,7 +1206,7 @@ namespace UnityGLTF
 
 			if(meshObj.boneWeights.Length != 0)
 			{
-				aJoint0 = ExportAccessor(SchemaExtensions.ExtractJointAndCopy(meshObj.boneWeights));
+				aJoint0 = ExportAccessor(SchemaExtensions.ExtractJointAndCopy(meshObj.boneWeights), SemanticProperties.JOINTS_0);
 				aWeight0 = ExportAccessor(SchemaExtensions.ExtractWeightAndCopy(meshObj.boneWeights));
 			}
 
@@ -1227,7 +1227,7 @@ namespace UnityGLTF
 				if (topology == MeshTopology.Triangles) SchemaExtensions.FlipTriangleFaces(indices);
 
 				primitive.Mode = GetDrawMode(topology);
-				primitive.Indices = ExportAccessor(indices, true);
+				primitive.Indices = ExportAccessor(indices, SemanticProperties.INDICES);
 
 				primitive.Attributes = new Dictionary<string, AccessorId>();
 				primitive.Attributes.Add(SemanticProperties.POSITION, aPosition);
@@ -1936,7 +1936,7 @@ namespace UnityGLTF
 			return id;
 		}
 		
-		private AccessorId ExportAccessor(int[] arr, bool isIndices = false)
+		private AccessorId ExportAccessor(int[] arr, string property)
 		{
 			uint count = (uint)arr.Length;
 
@@ -1945,9 +1945,12 @@ namespace UnityGLTF
 				throw new Exception("Accessors can not have a count of 0.");
 			}
 
+			bool isJoints = property == SemanticProperties.JOINTS_0;
+			bool isIndices = property == SemanticProperties.INDICES;
+
 			var accessor = new Accessor();
-			accessor.Count = count;
-			accessor.Type = GLTFAccessorAttributeType.SCALAR;
+			accessor.Count = isJoints ? count / 4 : count;
+			accessor.Type = isJoints ? GLTFAccessorAttributeType.VEC4 : GLTFAccessorAttributeType.SCALAR;
 
 			int min = arr[0];
 			int max = arr[0];
--- a/UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
+++ b/UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
@@ -1207,7 +1207,7 @@ namespace UnityGLTF
 			if(skin != null)
 			{
 				if (meshObj.bindposes.Length != 0)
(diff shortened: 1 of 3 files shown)


soraryu:
@owilliamailliwo couple questions to judge this vs my PR and which pieces to take from which:
- does this support multiple skeletons
- does this support multiple clips from an animator
- does this support actual keyframes or just baked animation (mine is currently "baked but optimized")
- does this have a timeline recorder clip (see unity usd for why this would be great) (mine doesn't yet)
- does this work with Google SceneViewer (I had to fix some things in gltf dll to make that work)

owilliamailliwo:
Hi @soraryu ,
does this support multiple skeletons	
**YES**
does this support multiple clips from an animator	  
**YES**
does this support actual keyframes or just baked animation (mine is currently "baked but optimized")	
**It supports keyframes animation.**
does this have a timeline recorder clip (see unity usd for why this would be great) (mine doesn't yet)	
**It will soon support this feature.**
does this work with Google SceneViewer (I had to fix some things in gltf dll to make that work)  
**I haven`t tested it with Google SceneViewer, but it works fine with babylon.js. It is fully compatible with the GLTF 2.0 standard.**

I will create a new pull request recently, which will support the compression of animation data and geometry properties.
Because we have multiple teams are developing game projects base on this plugin, we will quickly improve the functionality of the plugin
and share them out.

soraryu:
That's great to hear! I'll look into any differences that there may be between mine and yours and notice here.

About compatibility: note that the rest of Khronos' Unity code (currently) itself isn't fully spec compatible in some places unfortunately, and there are big differences in gltf support between babylon, three, filament, scene viewer try etc. Please try in all of them.

UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
```diff
@@ -637,7 +724,108 @@ private static bool IsPrimitive(GameObject gameObject)
...
 				&& gameObject.transform.localScale == Vector3.one
 				&& ContainsValidRenderer(gameObject);
+		}
+		
+		private void ExportSkin(string name, GameObject[] primitives, Node node)
+		{
+			foreach (var prim in primitives)
+			{
+				var smr = prim.GetComponent<SkinnedMeshRenderer>();
+				if (_existedSkins.TryGetValue(smr.rootBone.GetInstanceID(), out SkinId skinId))
```
CgGraphic: Shouldn't use the smr.rootbone's instanceid as the key, different skinnmeshrender'  may have same rootBone

UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
```diff
@@ -637,7 +724,108 @@ private static bool IsPrimitive(GameObject gameObject)
...
+					Root = _root
+				};
+
+				_root.Skins.Add(skin);
 
+				node.Skin = skinId;
+				
+				node.Mesh = ExportMesh(prim.name, new GameObject[] { prim });
+
+				_existedSkins.Add(smr.rootBone.GetInstanceID(), skinId);
```
CgGraphic: same here

UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
```diff
@@ -637,7 +724,108 @@ private static bool IsPrimitive(GameObject gameObject)
...
+							}
+						}
+						
+						nodeId = new NodeId
+						{
+							Id = _root.Nodes.Count,
+							Root = _root
+						};
+						
+						_root.Nodes.Add(boneNode);
```
CgGraphic: Generate GLTF nodes is not add to existedNode dict and the _root.Nodes. Would cause duplicated nodes when add gameobject.

It would be more clear to generate GLTF nodes for every gameobject first, then parse components like mesh/skin

UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
```diff
@@ -637,7 +724,108 @@ private static bool IsPrimitive(GameObject gameObject)
...
+
+				skin.Skeleton = rootBoneId;
+
+				skinId = new SkinId
+				{
+					Id = _root.Skins.Count,
+					Root = _root
+				};
+
+				_root.Skins.Add(skin);
```
CgGraphic: Same problem

UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
odweibo: > It would be more clear to generate GLTF nodes for every gameobject first, then parse components like mesh/skin

Hi, thanks very much! I've fixed the problems you found. I agree with you at this problem, but I don't want to make sweeping changes to the current code structure.

owilliamailliwo: > It would be more clear to generate GLTF nodes for every gameobject first, then parse components like mesh/skin

Hi, thanks very much! I've fixed the problems you found. I couldn't agree more with your suggestion, but I don't want to make sweeping changes to the current code structure.

UnityGLTF/Assets/UnityGLTF/Runtime/Scripts/GLTFSceneExporter.cs
owilliamailliwo: I have fixed this problem, I will commit it later. Thank you!

tufeixp:
This will not work for runtime export 'cause AnimationUtility was called, and disregard legacy Animation system.

hybridherbst:
@tufeixp that's correct; as outlined above a good improvement here would be the ability to record via a timeline clip (which would also work at runtime). Similar to how Unity-USD does that.

hybridherbst:
Closed – outdated. Animation export is in UnityGLTF since 2.x

PR #562, opened 2020-02-17
Change flip texture coordinates with flip texture

owilliamailliwo:
Better not to flip the texture coordinates. If uv* represents something else other than coordinates, we cannot simply perform 1 - uv*.y on it.


hybridherbst:
@pfcDorn @marwie we may want to look at this again, we've discussed a couple times if this would make sense. For roundtrips and trying to keep them as lossless as possible I still prefer the UV flip, but not sure in the general case. I don't think it's good to introduce a special option either though (and keep both code paths – I think we want one).

By now it encompasses a lot more than what's in this PR; e.g. UV animation and so on need to be handled as well.

pfcDorn:
@owilliamailliwo Have you a specific use case where the flipped uv's are an issue for you? And when yes, can you please describe it exactly :)

PR #569, opened 2020-03-20
Solve the problem that the exported root objects are not at the origin in world space

owilliamailliwo:
Replace "Export Selected" and "ExportGLB Selected" menus with "Export Selected", "ExportGLB Selected", "Export Selected With Root Transform" and "ExportGLB Selected With Root Transform", to solve the problem that the exported root objects are not at the origin in world space. You can still export objects with world transform with "Export Selected With Root Transform" and "ExportGLB Selected With Root Transform" menus.


hybridherbst:
Better late than never: we recently added batch export options, and "Reset" is one of the new batch export modes, which does exactly what you're describing here.

PR #576, opened 2020-04-12
Persistent store of Settings

owilliamailliwo:
Persistent Settings to asset so you don't have to set them every time you restart Unity.


hybridherbst:
Closed – outdated. Settings are persistently stored in the UnityGTLFSettings object.

Issue #577, opened 2020-04-14
TEXCOORD_4 is not an exception but an error of the model loading


EricBeetsOfficial:
A model loading failed with a 'TEXCOORD_4' attribute because in GLTFSceneImporter.cs:881 the GLTFLoadException is from UnityGLTF.GLTFLoadException (Exceptions.cs:35) and not from the GLTF.GLTFLoadException of the GLTFSerialization part (Exceptions.cs: 29).
Why is GLTFLoadException redefined in UnityGLTF ? Shouldn't you delete the class from UnityGLTF (and keep it in the GLTFSerialization part) ?

Thank you in advance.

PR #579, opened 2020-04-24
Fix the duplicated GLTFLoadException (generate a TEXCOORD_4 error instead of a warning)

EricBeetsOfficial-Opuscope:
As described in issue #499 and #577.
Delete a duplicated GLTFLoadException which emit an error instead of a warning when there is TEXCOORD_4 attribute.


hybridherbst:
Closed – the duplicate exception has been removed.

Issue #587, opened 2020-06-05
Failing to deserialize sparse accessors


dri-richard:
Our application reads glTF files from a 3rd party which would like to start using sparse accessors, but GLTFSerialization fails to deserialize them. These are files that pass validation.

Stepping through the code, AccessorSparse.Deserialize() immediately expects to find a PropertyName token, but given that a sparse data block is an object, e.g.:
`"sparse": {
 "count": 4,
 "values": {
 "bufferView": 2,
 },
 "indices": {
 "bufferView": 1,
 "componentType": 5123
 }
 }`
I'd expect it to read a StartObject token first.

I've experimented with this change, and deserialization now completes, but the scene importer fails with a NullReferenceException in ConstructMeshTargets() as it appears the targetAttributes are invalid.

I'm quite new to working with glTF files so I'm unable to proceed further, can anybody shed light on how I might debug this, or perhaps the sparse accessor support is incomplete and needs more work?

pfcDorn:
Fixed in latest release

PR #629, opened 2021-07-21
Fixes for sparse accessors and additional joints

jeoungjukim:
Hello, I am jeoungju kim, an employee of the graphic development team of Samsung Electronics Mobile Division.

This project has some fatal errors as well as issues with sparse accessors. So I share my edits.
The three modifications are:
1. Fix sparse accessor parsing error (#587)
2. Modified to support up to 2 joints
3. Modified to load accessors of Morph Targets with sparse accessors without bufferView


hybridherbst:
Thank you. Could you share a GLB file that breaks on the existing version and works with the fixes in this PR to verify?

PR #655, opened 2022-07-18
GLTFSceneImporter: handles the case where the mesh has 0 vertice.

glaporteavatar:
Avoid load exception.


hybridherbst:
Thank you! Added a similar change now on latest

PR #659, opened 2022-12-26
fix export humanoid translations (s/position/translation)

iamtommo:
bone translations were not being exported


hybridherbst:
Outdated – the codebase already contains this change.

Issue #376, opened 2019-03-14
lacking of some meshAttributes cause model import failure


tufeixp:
KeyNotFoundException: The given key was not present in the dictionary.
System.Collections.Generic.Dictionary`2[TKey,TValue].get_Item (TKey key) (at <ac823e2bb42b41bda67924a45a0173c3>:0)
UnityGLTF.GLTFSceneImporter.ConvertAccessorsToUnityTypes (UnityGLTF.MeshConstructionData meshConstructionData) (at Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs:1470)

Please refer to the attachment file.

tufeixp:
[ExpTargetGlbFile001.zip](https://github.com/KhronosGroup/UnityGLTF/files/2965546/ExpTargetGlbFile001.zip)

Zabr-din:
I am not sure if the repo is still supported, so there is a fork to solve this issue:

https://github.com/Zabr-din/UnityGLTF

pfcDorn:
Please update to the latest version and open a new issue if the problem persists. Thanks!

PR #665, opened 2023-03-03
Changed MeshPrimitive deserializer to ignore extras property

Zabr-din:
Fixes KhronosGroup/UnityGLTF#376


hybridherbst:
Should be fixed already, closing.

PR #676, opened 2023-11-22
Fix keyframe filtering

wschnepp:
Small fix pr that solves a problem with exporting animations from unity to gltf.
In particular, it changes how the `AnimationData.Track` class will handle identical consecutive samples.
The change is also explained in a comment in the code, so i am pasting that in here, since its more than enough explanation.
```
// As a memory optimization we want to be able to skip identical samples.
// But, we cannot always skip samples when they are identical to the previous one - otherwise cases like this break:
// - First assume an object is invisible at first (by having a scale of (0,0,0))
// - At some point in time, it is instantaneously set "visible" by updating its scale from (0,0,0) to (1,1,1)
// If we simply skip identical samples on insert, instead of a instantaneous
// visibility/scale changes we get a linearly interpolated scale change because only two samples will be recorded:
// - one (0,0,0) at the start of time
// - (1,1,1) at the time of the visibility change
// What we want to get is
// - one sample with (0,0,0) at the start,
// - one with the same value right before the instantaneous change,
// - and then at the time of the change, we need a sample with (1,1,1)
// With this setup, now the linear interpolation only has an effect in the very short duration between the last two samples and we get the animation we want.

// How do we achieve both?
// Always sample & record and then on adding the next sample(s) we check
// if the *last two* samples were identical to the current sample.
// If that is the case we can remove/overwrite the middle sample with the new value.
```

### DIFF
--- a/Runtime/Scripts/Timeline/GLTFRecorder.cs
+++ b/Runtime/Scripts/Timeline/GLTFRecorder.cs
@@ -229,23 +229,43 @@ namespace UnityGLTF.Timeline
 				private AnimationData tr;
 				private ExportPlan plan;
 				private Dictionary<double, object> samples;
-				private object lastSample;
+				private Tuple<double, object> lastSample = null;
+				private Tuple<double, object> secondToLastSample = null;
 				
 				public Track(AnimationData tr, ExportPlan plan, double time)
 				{
 					this.tr = tr;
 					this.plan = plan;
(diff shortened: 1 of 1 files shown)


hybridherbst:
Hi and thanks! @pfcDorn can you take a look please?

hybridherbst:
Merged in latest release, not sure why GitHub didn't pick that up – please let me know if that resolves your issue!

PR #684, opened 2024-02-02
Fix: Don't write "buffers" property in json if buffers are all empty

pfcDorn:
Added check if any buffer has bytes, otherwise don't write buffers property to json.


hybridherbst:
@pfcDorn would you please also add a test model that reproduces the error (before merging this PR) to the test models repository? Thank you!

Issue #681, opened 2024-02-01
HumanoidSetup needs to check "isHuman"


FujiSunflower:
# Before
If HumanDescription is wrong, the avatar is not human but generic as "isValid".
https://github.com/KhronosGroup/UnityGLTF/blob/d9ed104fd3ecc10265c7836484cd68e34cbedc13/Editor/Scripts/Internal/HumanoidSetup.cs#L25

# Aftar
So we will fix it like this.
`if (!avatar.isValid || !avatar.isHuman)`

# References
https://docs.unity3d.com/ScriptReference/Avatar-isHuman.html
https://docs.unity3d.com/ScriptReference/Avatar-isValid.html

pfcDorn:
i have fixed it with 5363224499393edf03d7177a40139618a5907273
Thx!

PR #685, opened 2024-02-02
Fix: issue #681 - humanoidsetup need to check isHuman

pfcDorn:
Added check "avatar.isHuman"  when building Human Avatar, to ensure it's valid.


hybridherbst:
Please make sure to add a test model that reproduces the issue before this PR would be merged!

pfcDorn:
After some more investigation, i would say the additional isHuman check is not needed here. BuildHumanAvatar already restrict it to a HumanAvatar and not a GenericAvatar. 
image
Test Model: 
[MetaAvatar.zip](https://github.com/KhronosGroup/UnityGLTF/files/14415778/MetaAvatar.zip)

So i would ignore this PR

PR #686, opened 2024-02-02
Feature: blendshape import > frame weight multiplier option

pfcDorn:
New option for importing blendshapes:
- Adjustable frame weight multiplier (default is 1x)

When using other animation on a gltf model with blendshapes, it can be helpful to adjust the range for the blendshapes (eg. 0...1 or 0..100)


hybridherbst:
Please always use `nameof` instead of string access, where possible!
image

PR #691, opened 2024-02-07
Added current loaded GLTF filename to exceptions and logs

pfcDorn:
Especially when Unity is importing a lot of gltf files, it can be hard to find out which files exactly has thrown or logged an error.


hybridherbst:
This PR breaks the provided logger:
image

The `Debug.` calls go to the provided ILogging instance (by default Unity's console logging but can be overriden), now they are directly on Unity's logging again.

PR #694, opened 2024-02-13
Import: Normalizing for quantized data fix

pfcDorn:
Added missing Accessor.Normalized checks, when accessor data get builded.
Before, some quantized accessor data gets always normalized (e.g. positions).


hybridherbst:
Are you sure about this change? I think it requires a bit more explanation on what is going on here.

image

PR #708, opened 2024-03-06
Export guard / layer mask fix

dbb789:
This PR has two relatively simple changes;

IGLTFSceneExportGuard - this interface allows a MonoBehaviour on a GameObject to mark a portion of the transform hierarchy as unexportable, which allows us to easily indicate that portions of a scene are for business logic only and not visible objects, without having to create new layers and/or adjust tags. I've also included GLTFSceneExportGuard which is an empty MonoBehaviour purely for convenience when using this feature within the editor.

On line 970, there's a check for _exportLayerMask >= 0. I assume this was written with the assumption that an empty layer mask would be nonsensical and the exporter should just default to all layers. However, if layer 31 is enabled, this sets the sign bit on the underlying int, which will make the layer mask negative, and it'll therefore fail this check. I've switched this to _exportLayerMask != 0.

Both of these changes have been tested by us internally. Please take a look - thanks!

### DIFF
--- a/Runtime/Scripts/GLTFSceneExportGuard.cs
+++ b/Runtime/Scripts/GLTFSceneExportGuard.cs
@@ -1,0 +1,9 @@
+using UnityEngine;
+
+namespace UnityGLTF
+{
+    // Empty implementation of IGLTFSceneExportGuard for use with editor.
+	public class GLTFSceneExportGuard : MonoBehaviour, IGLTFSceneExportGuard
+	{
+    }
+}
--- a/Runtime/Scripts/GLTFSceneExportGuard.cs
+++ b/Runtime/Scripts/GLTFSceneExportGuard.cs
@@ -2,8 +2,8 @@ using UnityEngine;
 
 namespace UnityGLTF
 {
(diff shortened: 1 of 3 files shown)


hybridherbst:
Hey, and thanks for the PR!

There are multiple ways already to exclude objects from export (via enabled state if wanted, via layers, via export callbacks).
I'm not sure if adding another core mechanism is the right approach here – but the same functionality can totally be achieved with a simple export plugin which would be an excellent small sample to ship with the package.

@pfcDorn can you take a look at making such an example export plugin and putting them into the Samples~ folder so it can be installed from PackMan? And also at the second half of this PR regarding the incorrect LayerMask check? Thanks!

dbb789:
We've been playing around with import/export plugins but couldn't see an obvious and/or approved way to selectively chop nodes out based on our own criteria. If there's a way to do that then please do let us know as that would be perfect - thanks!

pfcDorn:
Hey, with the next release, we will have a ShouldNodeExport callback for Plugins for such functionality.  You can also find a code snippet in the readme then.

PR #716, opened 2024-03-19
Removed synchronous wait in GLB json loading causing spikes

umeraamir-arthur:
The synchronous wait here was causing the main thread to wait until the JSON is loaded completely which was highly dependent on the device and was varying depending on it.

This new solution is also waiting synchronously but main thread doesn't wait for it.

### DIFF
--- a/Runtime/Scripts/GLTFSceneImporter.cs
+++ b/Runtime/Scripts/GLTFSceneImporter.cs
@@ -597,7 +597,12 @@ namespace UnityGLTF
 				Thread loadThread = new Thread(() => _gltfStream.Stream = dataLoader2.LoadStream(jsonFilePath));
 				loadThread.Priority = ThreadPriority.Highest;
 				loadThread.Start();
-				RunCoroutineSync(WaitUntilEnum(new WaitUntil(() => !loadThread.IsAlive)));
+				while (loadThread.IsAlive)
+				{
+					Debug.Log("GLB json data is loading async...");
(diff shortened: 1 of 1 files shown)


umeraamir-arthur:
@pfcDorn what do you say about this solution?

pfcDorn:
why not using await Task.Run( () => { ... } ); for loading and parsing the stream. Is much simpler

umeraamir-arthur:
So technically this whole code

`Thread parseJsonThread = new Thread(() => GLTFParser.ParseJson(_gltfStream.Stream, out _gltfRoot, _gltfStream.StartPosition));
parseJsonThread.Priority = ThreadPriority.Highest;
parseJsonThread.Start();
while (parseJsonThread.IsAlive)
{
	Debug.Log("GLB json data is loading async...");
	await Task.Delay(25);
}
Debug.Log("GLB json data is loaded.");`

will become like this?
`await Task.Run(() =>
{
	GLTFParser.ParseJson(_gltfStream.Stream, out _gltfRoot, _gltfStream.StartPosition);
});`

pfcDorn:
right! :)

umeraamir-arthur:
Great! What about thread priority?
`parseJsonThread.Priority = ThreadPriority.Highest;`
Will it affect?

pfcDorn:
it don't think it's important anymore. Every device will have enough CPU cores today :)

PR #717, opened 2024-03-19
Divided animations loading into multiple steps

umeraamir-arthur:
Animations were loading in a single step causing thousands of calls to the KeyFrames creation related functions in a single frame until an animation is loaded.

Most of the unoptimized GLBs have thousands of keyframes and to load them in a single frame was causing CPU spikes because of iterations and calling of functions for each keyframe.

Using Task.Yield(); here was a benefit because there are possibilities when there are less keyframes in a single channel. Task.Yield() doesn't always shift you in next frame. It just adds a break and if next iteration can be done on the same frame then it'll not wait for next frame to load small channels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this package will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.10.0-rc.2] - 2024-03-05
+- fix: add `com.unity.mathematics` as required dependency
+- fix: added null check in mesh data preparation code to prevent exception for files without nodes (e.g. just materials)
+
 ## [2.10.0-rc] - 2024-03-04
 - fix: GLTFRecorder issue where resulting animation would have linear interpolation for cases where a jump was expected
 - fix: GLTFSettings toolbar active index correctly stored in session now
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "org.khronos.unitygltf",
   "displayName": "UnityGLTF",
-  "version": "2.10.0-rc",
+  "version": "2.10.0-rc.2",
   "unity": "2020.3",
   "description": "Import and export glTF 2.0 assets, both in the editor and at runtime.",
   "repository": {


hybridherbst:
Thanks for the PR!

Please try to include test model(s) in issues or PRs so that we can verify the results and ideally add those models to the test file sets. If the files are under a permissive license that's even better (but not required).

Can you add one or more test models that show the difference here?

umeraamir-arthur:
Hi @hybridherbst 
Sorry I happened to miss the notification somehow otherwise I was waiting for the feedback.

Here is a GLB which has a lot of keyframes. Even Unity struggles to visualize them at edit time. We had a lot of GLBs even downloaded from paid websites but having hundreds of keyframes causing big lags when loading them.

[animated_flag_free.zip](https://github.com/KhronosGroup/UnityGLTF/files/14718065/animated_flag_free.zip)

pfcDorn:
hm... But this makes it also slower when there is a lot of animation channels. Each Task.Yield will cost basically 1 frame. I would prefer to add a time measurement to this and only call Task.Yield after a specific threshold of 5 ms or so

umeraamir-arthur:
Up until my testing results, Task.Yield() is not equals to 1 frame so an animation which was loading on 1 frame is now divided into 3-4 frames even there are 10-15 channels because of the nature of Task.Yield().

umeraamir-arthur:
Yes I also used to think the same but I tested printing [Time.frameCount](https://docs.unity3d.com/ScriptReference/Time-frameCount.html) after each for loop before Task.Yield and the results showed same values more often when channels had less keyframes (showing that they executed on same frame).
I also tried Unity profiler with deep profiling enabled. It also resulted same that keyframes creation function was called on way lesser frames then there are channels on the animations.

pfcDorn:
hm...that would be a strange behaviour which i can't reproduce: 
image
image

Do not wonder why the first both are in frameCount 1. The Yield will be continued between Update and LateUpdate.

pfcDorn:
interesseting....i don't get the same result. Which Unity Version are you using?

umeraamir-arthur:
2021.3.15f1 xD

umeraamir-arthur:
I just quickly tested it with 2021.3.35f1 and 2021.3.15f1 but find no luck reproducing it in a fresh project.

Maybe it is some project settings or something which is causing this behavior. In any case, if it is not universal I think we should use other thing..... What about `await Task.Delay(5);`

umeraamir-arthur:
image

It seems to be a bad idea to use `await Task.Delay(5);`

Most generic solution might be using Task.Yield() but after a certain interval and not after every channel..... What do you say about that??

pfcDorn:
for runtime we actually have already a solution ( i totally forgot about it ). Just using 
`await YieldOnTimeoutAndThrowOnLowMemory();`
When using the GLTFComponent, it will then using the AsyncCoroutineHelper component, which manage the budget per frame:
image

umeraamir-arthur:
Seems like a good idea..... So I can add it instead of Task.Yield() on same place i.e. after every channel for loop it'll check time taken for the frame. Right?

pfcDorn:
yeah...should be absolutly fine :)

umeraamir-arthur:
I'll test and update my branch in the PR.....

Should I force revert my previous commit or I can add 1 commit above the previous one?

umeraamir-arthur:
@pfcDorn I just reverted the commit and added a new one on top of it... It seems like a good improvement since it is better in many terms.... **"UmerAamir/optimization-animation-loading" branch has this new commit but the pull request is closed**....
@hybridherbst

PR #727, opened 2024-04-05
Fixed #726 - Tangent issues with Draco compression

### LINKED ISSUE #726: Mesh with tangents has validation errors and breaking compression
Model exported with UnityGLTF has thousands of tangent validation error

draco compression fails with access out of memory exception for these models

[model.zip](https://github.com/KhronosGroup/UnityGLTF/files/14873695/model.zip)

image

pfcDorn:
To fix the issue, 
we now create a proper Attribute Id Map for Draco. Before it was only created for weights and joints. For Tangents this seems not to be enough. I also added all other Mesh Attribute Ids with this change.

Test File:
[Preview.zip](https://github.com/KhronosGroup/UnityGLTF/files/14882061/Preview.zip)

### DIFF
--- a/Runtime/Scripts/SceneImporter/ImporterMeshes.cs
+++ b/Runtime/Scripts/SceneImporter/ImporterMeshes.cs
@@ -265,8 +265,30 @@ namespace UnityGLTF
 						decodeSettings |= DecodeSettings.RequireTangents;
 					if (firstPrim != null && firstPrim.Targets != null)
 						decodeSettings |= DecodeSettings.ForceUnityVertexLayout;
+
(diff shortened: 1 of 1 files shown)


marwie:
Issue: https://github.com/KhronosGroup/UnityGLTF/issues/726

hybridherbst:
@pfcDorn do we need to guard this with an ifdef against the older draco package? (which is in other places still supported as optional dependencies).

pfcDorn:
i don't think so, is already in an ifdef because of API changes. The only reason for another ifdef could be for an error message that tangents are not supported by older drago versions

PR #744, opened 2024-06-27
fixed TMPro export with multiple font weights (invalid texture hash)

Vaso64:
I had trouble exporting objects with TMPro text, containing multiple font weights (bold, italic etc.)
The UnityGLTF correctly exports all font atlas materials, but their textures are all the same.
After closer inspection, it seems that these textures do not have a valid `Texture.imageContentsHash`, which leads to exporter to just re-use the previously cached atlas texture, even if they are different.
I made it so in case of `!Texture.imageContentsHash.isValid`, the `UniqueTexture.GetHashCode()` fallbacks to using `Texture.GetHashCode()`.

Example:
fix_demo


hybridherbst:
Thank you!

PR #745, opened 2024-06-28
Unity 2020.3.30f1 Exporting a prefab from Project will result in missing animations.

ArminRoiderKubat:
gameObject.activeInHierarchy will return false if the GameObject is a prefab and not instantiated in the scene.

### DIFF
--- a/Runtime/Scripts/SceneExporter/ExporterAnimation.cs
+++ b/Runtime/Scripts/SceneExporter/ExporterAnimation.cs
@@ -779,7 +779,7 @@ namespace UnityGLTF
 					Transform targetTr = target.Length > 0 ? transform.Find(target) : transform;
 					int newTargetId = targetTr ? GetTransformIndex(targetTr) : -1;
 
-					var targetTrShouldNotBeExported = targetTr && !targetTr.gameObject.activeInHierarchy && !settings.ExportDisabledGameObjects;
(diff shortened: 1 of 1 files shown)


pfcDorn:
Seems to work with Unity 2021.3.+ without these changes.
@hybridherbst, whats your opinion? it would only benefit the compatibility support for 2020.3

hybridherbst:
I remember we had to do changes in the past to make that usecase work (exporting from Project), despite `activeInHierarchy` being false.

@ArminRoiderKubat could you please provide reproduction steps and a reproduction .unitypackage? Thanks

ArminRoiderKubat:
Here is the package and the exported glTF files.
[PR745.zip](https://github.com/user-attachments/files/16941337/PR745.zip)

```CubeScene.gltf``` will contain animations.
```CubeProject.gltf``` has no animations.
Both are exported from the same prefab. The first was exported by selecting the Prefab instance in the Hierarchy window in a open scene. The other by selecting it in the Project tab.

Steps:
- Open the ```SampleScene```, select the ```CubeScene``` GameObject and "Export selected as glTF".
- Select the Prefab ```CubeProject``` and "Export selected as glTF".
- Compare both files. The second export does not contain animations.

hybridherbst:
We determined this issue only affects Unity 2020.3, which isn't supported by either Unity nor UnityGLTF anymore. Thus, we're closing this PR to not add legacy code to the current codebase. Hope you understand!

PR #767, opened 2024-08-26
Add "ShouldNodeExport" hook

andybak:
Example usage:

        public override bool ShouldNodeExport(GLTFSceneExporter exporter, GLTFRoot gltfRoot, Transform transform)
        {
            Type[] excludedTypes =
            {
                typeof(FooBar),
                typeof(BarBaz)
            };
            bool hasExcludedComponent = excludedTypes.Any(t => transform.GetComponent(t) != null);
            return !hasExcludedComponent;
        }


andybak:
I've just hit a need for a similar method for Import. Does that belong in the same PR?

Is there anything missing that would help move this PR forward? We're a branch including this PR in production now.

hybridherbst:
We're looking at the PR again and looks like it will be included with the next version :) We did add a big warning to the Readme that using this hook requires understanding what can go wrong because it's rather easy to shoot yourself in the foot (e.g. skipping export of a few bones in a hierarchy).

hybridherbst:
Would be great if you take a look at that then and yes, feel free to make a new PR for import functionality!

hybridherbst:
This is now merged and released in 2.15 (not sure why GitHub doesn't notice that in this PR) 🎉

PR #772, opened 2024-08-29
Feature: Resource Deduplication on import

pfcDorn:
New Import Options: Resource Deduplication for Meshes and Textures

Searches for identical Meshes/Textures on import and remaps/replaces the corresponding id's.
For referenced images by URIs, only the URI will be taken for comparision, not the image data inside the files.


hybridherbst:
Merged into dev

PR #774, opened 2024-09-04
added gltf Asset informations to Importer Inspector

pfcDorn:
Enhanced the importer inspector to show the gltf asset informations: 

image
image

The informations will be stored in the importer as a multiline string from the new ToString methods in the gltf "Asset" class.


hybridherbst:
cc @marwie

marwie:
Which glTF files did you use here? Do you omit the `extras` info in the first screenshot when there's Copyright information in the glTF? How would it look with this glb? https://cloud.needle.tools/api/v1/public/9283996/105938669/mustang_sunglasses_-_mu_1683.glb

pfcDorn:
Was two different files. One without containing the extras.

Here's your file: 

image

PR #826, opened 2025-03-25
Basic HDRP export support

krisrok:
This adds basic export support for some of HDRP's standard shaders:
- Normal map 
- Metallic/smoothness via scalars or mask map
- Occlusion via mask map
- Emission via HDR and LDR+intensity workflows and emission map

To support the remapping of the mask map values, the channel swap shaders now support min/max values per used channel.
`smoothnessMultiplier` is now `smoothnessRangeMax` and `smoothnessRangeMin` has a default value of 0. Newly introduced min/max ranges also default to 0/1. This way, it should work exactly like before for material exports not using the ranges.

There's much to be desired for full support, even in the added functionality: UV sets, multiplicative emission, emission exposure weight come to mind. But for my use cases it works just fine and produces good looking results.

### DIFF
--- a/Runtime/Resources/MetalGlossChannelSwap.shader
+++ b/Runtime/Resources/MetalGlossChannelSwap.shader
@@ -4,6 +4,9 @@
 	{
 		_MainTex ("Texture", 2D) = "white" {}
 		_SmoothnessMultiplier ("Smoothness Multiplier", float) = 1
+		_SmoothnessMultiplier2 ("Smoothness Multiplier2", float) = 0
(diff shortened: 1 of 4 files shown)


hybridherbst:
Thanks for the PR! We'll take a look. Are you planning on adding the remaining changes that are needed to integrate it, e.g. into the animation system? (e.g. different treatment for emissive color animation and so on)

hybridherbst:
We'll aim to include this PR in 2.16 in the next week(s) :)

krisrok:
Cool! Regarding your question about extending this any further: Not at the moment, my needs are fulfilled for now :)

PR #829, opened 2025-04-01
Audio emitter implementation in the KHR_Interactivity branch

dqmachine:
I forked and branched off of the KHR_Interactivity branch to implement my audio changes since I need the changes in the interactivity branch and the interactivity PR is still open.

This supports KHR_Audio_Emitter.  Resused parts of the audio example already in project where I could.

### DIFF
--- a/Editor/Scripts/Interactivity/VisualScriptingExport/ExportGraph.cs
+++ b/Editor/Scripts/Interactivity/VisualScriptingExport/ExportGraph.cs
@@ -7,6 +7,9 @@ using UnityGLTF.Interactivity.VisualScripting.Export;
 
 namespace UnityGLTF.Interactivity
 { 
+    /// <summary>
+    /// Graph containing exporter nodes for use by the interactivity and audio contexts
+    /// </summary>
     public class ExportGraph
     {
         public GameObject gameObject = null;
--- a/Editor/Scripts/Interactivity/VisualScriptingExport/ExportGraph.cs
+++ b/Editor/Scripts/Interactivity/VisualScriptingExport/ExportGraph.cs
@@ -1,0 +1,19 @@
+using System.Collections;
+using System.Collections.Generic;
+using Unity.VisualScripting;
+using UnityEngine;
+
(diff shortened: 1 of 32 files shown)


hybridherbst:
Oops, I didn't realize deleting the branch now that it's merged closes this PR automatically.
@dqmachine could you please rebase the PR on main and open it again? Thanks!

pfcDorn:
Hey,
I made some changes and implement (WIP) the KHR_audio_emitter Extension in a more generall way to use it with UnityGltf. 
You can take a look at the [KHR_audio_emitter PR](https://github.com/KhronosGroup/UnityGLTF/pull/835)

I take a look with your audio node implementations, i would not recommend to use any strings as an value input. Yes, it's currently in the interactivity extension (i should removed it), but it's not an official Gltf Interactivity Type. And i don't think you need it there. I would say the audiosource id/index should be enough as an Input, or declaring it in the config. So when you stop a audiosource with a specific id, all nodes which reference this emitter>audiosource id, should stop playing.

pfcDorn:
It's totally separate. It's just a plugin for the core UnityGltf. So it should work flawlessly with your interactivity implementation :)

hybridherbst:
I think the general approach for adding entirely new interactivity features should be:
- adding import/export support for a particular extension independent of interactivity (e.g. audio, video)
- wiring it up with interactivity features through Unit exporters

hybridherbst:
@dqmachine could you please clarify why you're now implementing a new GOOG_audio extension instead of basing on KHR_audio_emitter? Thanks!

eugenecys:
We're proposing a stripped down version of the audio extensions as we have some features that we are planning for release that will most likely happen before KHR_audio_emitter is ratified, and prefer to have something stable so we won't run into compatibility issues should the KHR_audio_emitter spec changes.

The plan is to remove this once KHR_audio_emitter and KHR_audio_graph is ratified or close to it. Since only Google is proposing this and also to make it clearer it's not meant to be or replace the main audio extensions, GOOG_audio makes more sense.

PR #847, opened 2025-05-01
Add async callbacks to GLTFImportPluginContext

eoineoineoin:
I had a use-case when implementing an import plugin and needed to call an async function. Attempting to do this from the synchronous callbacks would either trigger exceptions due to code which should only run on the main thread or would hang Unity.

This PR extends the GLTFImportContext to provide async versions of the callbacks; for backwards-compatibility, the base implementation calls the synchronous version, so any extant implementations of the GLTFImportPluginContext should not see any API changes due to this.

For my use-case, I just needed an async OnAfterImportNode(), but saw no reason that this shouldn't be supported for each callback.

### DIFF
--- a/Runtime/Scripts/GLTFSceneImporter.cs
+++ b/Runtime/Scripts/GLTFSceneImporter.cs
@@ -430,7 +430,7 @@ namespace UnityGLTF
 				// TODO check where the right place is to call OnBeforeImport as early as possible
 				foreach (var plugin in Context.Plugins)
 				{
-					plugin.OnBeforeImport();
+					await plugin.OnBeforeImportAsync();
 				}
 
 				if (_options.ThrowOnLowMemory)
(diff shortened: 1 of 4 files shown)


hybridherbst:
Thanks for the PR! Unfortunately it's not quite as easy as this – async code can't be called from editor-time importers, it must run synchronous to the best of my understanding.

For the same reason, there are already helper classes in place to do that, would it be possible for your use case to instead run your async method synchronous inside the existing ``OnAfterImportNode``?

The helpers are here:
https://github.com/KhronosGroup/UnityGLTF/blob/4da23d4f010d1bbdfec0e024bb9f2758516ecf5f/Editor/Scripts/Internal/AsyncHelpers.cs#L16

eoineoineoin:
Thanks for looking into this; wasn't aware of that AsyncHelpers at all - looks like it does the trick and works correctly; I think that's preferable to use, rather than making the plugins async. Think we can close this PR, but wonder if there's some way we could bring attention to that util, just for developer productivity?

PR #849, opened 2025-05-12
Interactivity Runtime Playback

Jmangles:
Migration of everything from https://github.com/magicleap/UnityGLTF-Interactivity to UnityGLTF itself.

Some extra work required to implement behaviour engine and wrapper component creation in ImporterPlugin/Context files before merging.

Supports every node in the current spec along with core pointers and most extension pointers. Includes unit tests for every node that can be run in Unity and automatically export as json for consumption in other engines.


Runtime/Scripts/Interactivity/Playback/Context/Export/InteractivityExportContext.cs
pfcDorn, line 1: Are this logs still necessary?

pfcDorn:
i get an error in Unity:
",,,NodeRegistry.cs(27,51): error CS0246: The type or namespace name 'DebugLog' could not be found (are you missing a using directive or an assembly reference?)"

pfcDorn:
and i would suggest to change the plugin names to "KHR_interactivity Playback Importer" and "KHR_interactivity Playback Exporter", just for cleareance  :)

pfcDorn:
Do you have an exactly reason why are you not using "OnAfterImportScene" , so you're required to add OnLoadingComplete? Between these both callbacks are mostly only memory disposables and cleanups

pfcDorn:
> > Do you have an exactly reason why are you not using "OnAfterImportScene" , so you're required to add OnLoadingComplete? Between these both callbacks are mostly only memory disposables and cleanups
> 
> I found that in OnAfterImportScene that scene.Extensions is null. I guess that's more of a bug to be fixed than a reason to add another callback.

OnAfterImportScene has the Scene not the Root as parameter. Don't get confused here, the Extensions of the Gltf are on the Root, not the scene. But you already get correctly the Root in your OnLoadComplete callback with _context.SceneImporter.Root. So just use the OnAfterImportScene with the same code as in OnLoadComplete.
image

pfcDorn:
I tried to import a Gltf Asset with interactivity, sadly... your BehaviourEngine is not serializable. Would be really great if you can make it serializable, so your implementation would also work with Gltf Assets. Or maybe save your imported Graph as an subasset, not sure.
Also the EventWrapper name is a bit confusing. Maybe BehaviourEnginePlayer or something like that would match better :)

robertdorn83:
hey, i was taking a look on the latest changes. your variable/set is using the wrong index for the input port. 
image
you need to use the "index" instead of "i" for the value input port. 
I already wrote alexey in the interactivity PR that this need to be more clear in the spec, what is meant with "index" here.

robertdorn83:
Hey,
do you plan to update runtime to the new Ref spec changes? :)

hybridherbst:
Hey @Jmangles, are you planning to update this PR with the latest spec changes, or have plans changed? All the best!

Jmangles:
I haven't kept up to date with the spec. What do you estimate the effort would be in order to update this to the current revision?

robertdorn83:
hard to tell, there's a new "ref" type, which could end up in a bit more work. the other changes are all small.

PR #879, opened 2025-07-13
fix(build issue): Missing namespace prevents successful build (DediuplicationCleanUp)

api-haus:
I guess this was removed by accident.

This is an actual issue with build, present in latest release version.


hybridherbst:
Sorry about that, during a namespace cleanup right before release we accidentally removed a using that should not have been removed.

Just pushed a fix, the package should be updated with 2.17.3 in a few minutes.

PR #885, opened 2025-08-31
Independent Axis Scaling on Import

roguesaloon:
This is a small change that is useful for our workflow (and likely others), to allow the GLTF import Scale Factor to be axis independent.

This allows an easy way to fix any assumptions that the modelling application might have made about co-ordinate space in a non-destructive way on import without changing any of the core logic around how GLTF calculates the co-ordinate space (Y-Up, Z-Forward). It also allows for much rarer scenarios where you might want to scale any axis with differing scale factors. 

Existing behaviour is maintained fully, with new independent scaling behaviour being exposed in the UI as a simple check box:

image

### DIFF
--- a/Editor/Scripts/GLTFImporter.cs
+++ b/Editor/Scripts/GLTFImporter.cs
@@ -85,7 +85,9 @@ namespace UnityGLTF
 
 	    [Tooltip("Turn this off to create an explicit GameObject for the glTF scene. A scene root will always be created if there's more than one root node.")]
         [SerializeField] internal bool _removeEmptyRootObjects = true;
-        [SerializeField] internal float _scaleFactor = 1.0f; 
+        [SerializeField] internal float _scaleFactor = 1.0f;
+        [SerializeField] internal bool _useAxisScaleFactor = false;
+        [SerializeField] internal Vector3 _axisScaleFactor = Vector3.one;
         [Tooltip("Reduces identical resources. e.g. when identical meshes are found, only one will be imported.")]
         [SerializeField] internal DeduplicateOptions _deduplicateResources = DeduplicateOptions.None;
         [SerializeField] internal int _maximumLod = 300;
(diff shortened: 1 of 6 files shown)


hybridherbst:
Thanks for the PR!

Can you share a file or files where this is necessary? Coordinate system changes should never be needed in a strict standard such as glTF, unless the exporting application does something wrong.

I believe for custom workflows this would be better to be in a plugin – shipping it as general feature has far to big risk for people messing up their pipeline.

roguesaloon:
> Thanks for the PR!
> 
> Can you share a file or files where this is necessary? Coordinate system changes should never be needed in a strict standard such as glTF, unless the exporting application does something wrong.
> 
> I believe for custom workflows this would be better to be in a plugin – shipping it as general feature has far to big risk for people messing up their pipeline.

It's probably less about the application (blender in our case), and more about assumptions made when creating the model. We were previously using UniGLTF which likely doesn't conform to the glTF standard as "strictly", we switched over to UnityGLTF to make use of the richer feature set (but still use UniGLTF for VRM).

UniGLTF has a setting (in unity preferences), that is on by default, `Default Invert axis: X`, so when we changed over to use UnityGLTF some of our models were now facing backwards (like they had been rotated 180 degrees around the X-axis) because they were expecting a different Z-Forward direction. I looked into implementing a workaround to to allow us to "fix" these models from the unity side on import (without needing to apply custom trasnforms directly to the game object in unity).

The more "correct" approach might have been to go into blender and apply the transforms there and then re-export with the strict glTF aligned co-ordinate system in mind, but we also found that a bit confusing to work with as blender uses a right handed co-ordinate system with Z-Up, Y-Forward. As we weren't even seeing entirely consistent results with the same model and different glTF importers (UniGLTF vs UnityGLTF), the implementation in this PR seemed like a fair solution to me (especially as it doesn't change existing default behaviour at all, and ultimately just adds flexibility).

As for creating this as an external import plugin for our workflow, that might be possible and I can take a look into it. However I'm not entirely sure that all the right hooks are exposed to be able to implement this cleanly as a plugin (Though I might be wrong on that).

hybridherbst:
UniGLTF is sadly not very spec-compliant (I have opened issues there in the past about that). Especially 180° rotation around Y is something that, sadly, some exporters get wrong; mostly the issue comes from the fact that some exporters approximate the glTF X-flip with 180° rotation around, which is wrong.

@pfcDorn could you make a mini example plugin for this kind of conversion on import, for demonstration?

PR #887, opened 2025-09-09
Fix: Add missing using UnityEngine directive

angree:
Fixes CS0246 compilation error in Unity 2022.3, Unity 2021.3 Added missing 'using UnityEngine;' for RuntimeInitializeOnLoadMethodAttribute


robertdorn83:
Hey, thanks for the fix!!

PR #900, opened 2026-03-02
Fix NullReferenceException for morph target accessors without BufferView

SmittyWerbenJJ:
Importing a glTF file with morph targets exported from Blender (confirmed also 
reproducible with files exported from Godot) throws a `NullReferenceException`:

```
Asset import failed, "Assets/.../Daisy_Import_File.glb" > NullReferenceException: 
Object reference not set to an instance of an object
UnityGLTF.GLTFSceneImporter.ConstructMeshTargetsPrepareBuffers (...)
  at ImporterMeshes.cs:786
```

The crash is on this line in `ConstructMeshTargetsPrepareBuffers`:
```csharp
bufferIdPair = targetAttribute.Value.Value.BufferView.Value.Buffer;
```

## Root Cause

The glTF 2.0 spec explicitly defines `accessor.bufferView` as optional.

**[Section 5.1.1 of the glTF 2.0 spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_accessor_bufferview)** (`accessor.bufferView`):
> "When undefined, the accessor **MUST** be initialized with zeros; `sparse` property or extensions **MAY** override zeros with actual values."

**[Section 3.6.2.3](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#sparse-accessors)** (Sparse Accessors) reinforces this:
> "When `accessor.bufferView` is undefined, the sparse accessor is initialized as an array of zeros of size (size of the accessor element) * (accessor.count) bytes."

This means `bufferView` AND `sparse` can both legally be null simultaneously. It simply means the accessor is all zeros.

For morph targets specifically, **[Section 3.7.2.2](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#morph-targets)** explicitly calls this out as an intended authoring pattern (emphasis mine):

> "Attributes present in the base mesh primitive but not included in a given morph target **MUST** retain their original values for the morph target."
>
> *Implementation Note: **This allows skipping zero-filled accessors** and implies that different morph targets may contain different sets of attributes.*

The existing code already handles the null `bufferView` case in the `sparse != null` 
branch (added for Draco support), but the `sparse == null` branch was missing the 
same guard and assumed `bufferView` would always be present.

## Fix

Added a null-check to the non-sparse branch, mirroring the existing Draco guard

### DIFF
--- a/Runtime/Scripts/SceneImporter/ImporterMeshes.cs
+++ b/Runtime/Scripts/SceneImporter/ImporterMeshes.cs
@@ -824,7 +824,7 @@ namespace UnityGLTF
 					BufferId bufferIdPair = null;
 					if (targetAttribute.Value.Value.Sparse == null)
 					{
-						if(targetAttribute.Value.Value == null) continue;
+						if(targetAttribute.Value.Value.BufferView == null) continue;
 						bufferIdPair = targetAttribute.Value.Value.BufferView.Value.Buffer;
 					}
 					else
--- a/Runtime/Scripts/SceneImporter/ImporterMeshes.cs
+++ b/Runtime/Scripts/SceneImporter/ImporterMeshes.cs
@@ -824,6 +824,7 @@ namespace UnityGLTF
 					BufferId bufferIdPair = null;
 					if (targetAttribute.Value.Value.Sparse == null)
 					{
+						if(targetAttribute.Value.Value == null) continue;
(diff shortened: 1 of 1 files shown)


robertdorn83:
hey, can you pls sign the [Contributor License Agreement](https://cla-assistant.io/KhronosGroup/UnityGLTF?pullRequest=900)? so i can merge it :)

PR #902, opened 2026-04-11
Unity 6.6: changes we used to get UnityGLTF working in a consumer project

CharlieHess:
# Unity 6.6 viability notes

This branch is not presented as a finished Unity 6.6 support release. It is a record of the changes we used in a consumer project to get UnityGLTF working again on Unity 6.6 alpha.

## What this fixes

- Replaces exporter-side `GetInstanceID()` usage that turns into hard failures on Unity 6.6 with `GetEntityId()`-based keys and labels.
- Updates editor asset creation code to the newer `AssetCreationEndAction` and `EntityId` APIs.
- Updates editor state and preview tracking code to use entity ids where Unity 6.6 no longer accepts instance ids.
- Forces the rough refraction renderer feature onto the RenderGraph-compatible path on newer Unity 6 releases where the older compatibility path is no longer viable.
- Replaces one Visual Scripting exporter `GUID.Generate()` call with `Guid.NewGuid()` to avoid an editor-side API mismatch we hit during the migration.

## Scope and limits

- This branch is intentionally narrow. It captures the concrete changes we needed for Unity 6.6 viability, not a broad compatibility audit across all UnityGLTF features.
- The validation evidence comes from the consumer project integration described above. We did not run a separate full UnityGLTF package test pass in this forked workspace.
- We left out unrelated package metadata and Unity-authored serialization noise from the consumer branch so the patch stays focused on the actual compatibility edits.

### DIFF
--- a/Editor/Scripts/GLTFExportMenu.cs
+++ b/Editor/Scripts/GLTFExportMenu.cs
@@ -622,14 +622,14 @@ namespace UnityGLTF
 
 			var importAction = ScriptableObject.CreateInstance<AdjustImporterAction>();
 			importAction.fileContent = content;
-			ProjectWindowUtil.StartNameEditingIfProjectWindowExists(0, importAction, filename, null, (string) null);
+			ProjectWindowUtil.StartNameEditingIfProjectWindowExists(EntityId.None, importAction, filename, null, (string) null);
 		}
 
 		// Based on DoCreateAssetWithContent.cs
-		private class AdjustImporterAction : EndNameEditAction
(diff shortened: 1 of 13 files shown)


hybridherbst:
Thanks for posting this! We've not yet looked into Unity 6.6 (focus is on stability in LTS releases), but it's good to get a heads-up about changes needed there.

robertdorn83:
Hey,
should be now fixed with version 2.19.4. A lot of these errors was already obsolete warnings in Unity 6.4. 
I keep this open, in case you find other issues in U6.6 :)

